PR #2, opened 2020-11-01
fix: use `.webmanifest` instead of `.json` for manifest

wtchnm:


antfu:
Thanks and good to know that. But since it's a generated file, maybe we can just change the extension instead of giving the options for users to opt-in.

wtchnm:
Done! :)

PR #3, opened 2020-11-30
fix: fixed path to sw.js

taigabrew:
If a page was loaded not from a root url (eg. `https://example.com/login`) script tries to load `sw.js` from current url like this – `https://example.com/login/sw.js`, not like this - `https://example.com/sw.js`.


antfu:
Released in v0.1.6. Thanks

taigabrew:
Somehow changes was not published 😢

Снимок экрана 2020-11-30 в 14 05 06

antfu:
Emmm, thanks. Should work now on 0.1.7.

PR #9, opened 2020-12-27
fix: default workbox.mode to "production"

### LINKED ISSUE #8: Minified output
Service worker and Workbox javascript files are not minified. Is there any option I should add or is it missing this feature?

azamat-sharapov:
`tsup` doesn't allow to include `process.env.NODE_ENV` piece (literally) in output. Instead, it writes value of it. This PR changes default of `workbox.mode` to `"production"`, because that's mostly desired, than `"development"`. 

`mode: "production"` results in minified output of service worker and workbox code.
Closes #8

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,7 +15,7 @@ export function VitePWA(options: Partial<VitePWAOptions> = {}): Plugin {
     swDest: `${outDir}/sw.js`,
     globDirectory: outDir,
     offlineGoogleAnalytics: false,
-    mode: process.env.NODE_ENV,
+    mode: process['env']['NODE_ENV'] || 'production',
   }
 
   const defaultManifest: Partial<ManifestOptions> = {
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,7 +15,7 @@ export function VitePWA(options: Partial<VitePWAOptions> = {}): Plugin {
     swDest: `${outDir}/sw.js`,
     globDirectory: outDir,
     offlineGoogleAnalytics: false,
-    mode: process.env.NODE_ENV || 'development',
+    mode: process.env.NODE_ENV,
   }
 
   const defaultManifest: Partial<ManifestOptions> = {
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -15,7 +15,7 @@ export function VitePWA(options: Partial<VitePWAOptions> = {}): Plugin {
     swDest: `${outDir}/sw.js`,
     globDirectory: outDir,
     offlineGoogleAnalytics: false,
-    mode: process.env.NODE_ENV || 'development',
+    mode: process.env.NODE_ENV,
```
antfu, line 18:
```suggestion
    mode: process.env.NODE_ENV || 'production',
```

Why not just default this to `production` while users can still manually change it? I don't see it's related to tsup tho.

azamat-sharapov: > I don't see it's related to tsup tho

When you run [`build` script](https://github.com/antfu/vite-plugin-pwa/blob/master/package.json#L21), it calls `tsup`, which [replaces](https://github.com/egoist/tsup/blob/0b80615d053148ab907f41c402d09ec1eafc5e94/src/index.ts#L121) `process.env` variables with their value. This is how it's related. Ideally, you want `mode: process.env.NODE_ENV` in your `dist/index.js` instead of `mode: "development/production"`.

> Why not just default this to production

Hardcoded value is less useful than dynamic `process.env.NODE_ENV`, because you won't lose [`tsup`'s ability to control it](https://github.com/egoist/tsup/blob/0b80615d053148ab907f41c402d09ec1eafc5e94/src/index.ts#L93-L94).

antfu: Thanks for the clarification! I didn't notice that `tsup` will replace the env. Then I am thinking to have these to bypass the replacing

```js
mode: process['env']['NODE_ENV'] || 'production',
```

the `process.env.NODE_ENV` is meant to be the env from Vite env, not tsup.

WDTY?

azamat-sharapov: That's good idea, agreed. Hopefully `tsup` won't catch that syntax in the future as well, because if it does, `mode` becomes `"development"` again :)

Pushed changes.

Issue #4, opened 2020-12-06
Support "base" path from vite config


ubermanu:
Hey,

When building static assets from VitePWA, the paths to the service worker and the manifest is hardcoded (at root level)
Would it be possible to preprend them with the "base" config from vite?

antfu:
Would you mind to open a PR for it? Thanks.

antfu:
#13

PR #13, opened 2021-01-12
fix: service worker register path & use config build base (#4)

hannoeru:
In latest version, `serviceWorker.register()` has a wrong path.
`swDest: resolve(root, ${outDir}/sw.js),` has full path including root.
`workbox!.swDest.replace(outDir, '')` only replaced outDir.

add config build base (#4) in path and fix path in `manifest.webmanifest`.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,6 +14,7 @@ export function VitePWA(options: Partial<VitePWAOptions> = {}): Plugin {
   return {
     name: 'vite-plugin-pwa',
     enforce: 'post',
+    apply: 'build',
     configResolved(config) {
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -14,6 +14,7 @@ export function VitePWA(options: Partial<VitePWAOptions> = {}): Plugin {
   return {
     name: 'vite-plugin-pwa',
     enforce: 'post',
+    apply: 'build',
```
antfu, line 17: That's good to know! 👍

antfu:
Awesome! Thanks.

PR #14, opened 2021-01-15
types(manifest): add missing scope and orientation

posva:
I found them missing but I directly edited the types files hoping it's enough 😬


antfu:
Thanks!

PR #17, opened 2021-01-19
Option for registerServiceWorker (inline: boolean)

domsen123:
No inline javascript.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -29,6 +29,36 @@ importers:
       typescript: ^4.1.3
       vite: ^2.0.0-beta.23
       workbox-build: ^6.0.2
+  ../../../../../../../projects/forks/vite-plugin-pwa:
+    dependencies:
+      debug: 4.3.2
+      fast-glob: 3.2.4
+      pretty-bytes: 5.5.0
+      workbox-build: 6.0.2
+    devDependencies:
+      '@antfu/eslint-config': 0.4.3_eslint@7.17.0+typescript@4.1.3
+      '@types/debug': 4.1.5
+      '@types/workbox-build': 5.0.0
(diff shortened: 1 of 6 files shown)


pnpm-lock.yaml
```diff
@@ -29,6 +29,36 @@ importers:
       typescript: ^4.1.3
       vite: ^2.0.0-beta.23
       workbox-build: ^6.0.2
+  ../../../../../../../projects/forks/vite-plugin-pwa:
```
antfu, line 32: You probably want to undo this :P

src/config.ts
```diff
@@ -13,6 +13,7 @@ export function resolveOptions(options: Partial<VitePWAOptions>, viteConfig: Res
   const {
     srcDir = 'public',
     outDir = viteConfig.build.outDir || 'dist',
+    inline = true,
```
antfu, line 16: Maybe `inlineScript` could be more clear, what do you think?

src/html.ts
```diff
@@ -3,17 +3,32 @@ import { ResolvedVitePWAOptions } from './types'
...
+      }
+    </script>
+  </head>`
+    );
+  } else {
+    return html.replace(
+      '</head>',
+      `
+  <link rel="manifest" href="${join(basePath, 'manifest.webmanifest')}">
+  <script src="registerServiceWorker.js"></script>
```
antfu, line 30:
```suggestion
  <script src="registerServiceWorker.js"></script>
```

We shall respect the `base` option

pnpm-lock.yaml
domsen123: done :P

src/config.ts
domsen123: you are right - fixed this

src/html.ts
domsen123: done

PR #18, opened 2021-01-26
feat: new minify option

Dimitris-Toulis:
Minify the generated manifest to save some bytes

### DIFF
--- a/src/html.ts
+++ b/src/html.ts
@@ -3,7 +3,8 @@ import { FILE_MANIFEST, FILE_SW_REGISTER } from './constants'
 import { ResolvedVitePWAOptions } from './types'
 
 export function generateSWRegister(options: ResolvedVitePWAOptions) {
-  return `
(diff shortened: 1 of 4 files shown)


antfu:
I think we could simply have `minify` as the options name

src/html.ts
```diff
@@ -3,7 +3,8 @@ import { FILE_MANIFEST, FILE_SW_REGISTER } from './constants'
 import { ResolvedVitePWAOptions } from './types'
 
 export function generateSWRegister(options: ResolvedVitePWAOptions) {
-  return `
+  return options.minify ? `"serviceWorker"in navigator&&window.addEventListener("load",()=>{navigator.serviceWorker.register("${join(options.basePath, options.filename)}",{scope:"${options.scope}"})});` : 
+`
```
antfu, line 7: Don't think it's necessary here, can you omit this change?

Dimitris-Toulis: Done!

antfu:
Thanks

PR #21, opened 2021-02-04
fix: ensure using `base` does not require additional configuration

ElMassimo:
[relative]: https://developer.mozilla.org/en-US/docs/Web/Manifest/start_url#relative_url

### Description 📖 

This pull request makes a few changes to ensure that a Vite app using `base` doesn't require additional configuration in <kbd>vite-plugin-pwa</kbd>

### Changes ✏️ 

- Changed `start_url` to be [relative], so that when using `base` the `start_url` is resolved correctly.
- Removed a leading slash from the navigation fallback so that it matches the exact name that is rendered in the manifest, fixing an error when the service worker is first installed:
```
  WorkboxError.js:28 Uncaught (in promise) non-precached-url: non-precached-url :: [{"url":"/index.html"}]
```

### Notes 🗒️ 

Initially I thought about turning `cachePreset` into a function that receives `basePath`, and use:

```js
urlPattern: basePath,
```

but after investigating, it turns out that the first rule copied from <kbd>next-pwa</kbd> is not really used.

Since <kbd>vite-plugin-pwa</kbd> adds `index.html` to `navigateFallback`, the precaching rule will always match first, and that rule will never be used.

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -3,23 +3,6 @@
 import { RuntimeCachingEntry } from 'workbox-build'
 
 export const cachePreset: RuntimeCachingEntry[] = [
-  // if you are customizing your runtime cache rules, please note that the
-  // first item in the runtime cache configuration array MUST be "start-url"
-  {
-    // MUST be the same as "start_url" in manifest.json
-    urlPattern: '/',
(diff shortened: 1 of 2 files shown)


antfu:
Thanks for the high-quality PR 😄👍

PR #25, opened 2021-02-19
fix: html link error url when use CDN.

chengpeiquan:
Hi, Anthony. When I set the config option [base](https://vitejs.dev/config/#base) as CDN like 'https://cdn.jsdelivr.net/gh/chengpeiquan/chengpeiquan.com@gh-pages/' in `vite.config.ts`, I found a problem:

Run build on my project, then, in `/dist` folder, all `html` files was added these codes:

```html
<link rel="manifest" href="/https:/cdn.jsdelivr.net/gh/chengpeiquan/chengpeiquan.com@gh-pages/manifest.webmanifest">

<script src="/https:/cdn.jsdelivr.net/gh/chengpeiquan/chengpeiquan.com@gh-pages/registerSW.js"></script>
```

They are start with `/https....`, so, in the browser they are 404.

I found the cause of the problem in [html.ts](https://github.com/antfu/vite-plugin-pwa/blob/master/src/html.ts#L5-L7)

Now, this PR already fixed it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,4 +12,4 @@ jobs:
       - uses: actions/checkout@v2
       - run: npx conventional-github-releaser -p angular
         env:
-          CONVENTIONAL_GITHUB_RELEASER_TOKEN: ${{secrets.GITHUB_TOKEN}}
+          CONVENTIONAL_GITHUB_RELEASER_TOKEN: ${{secrets.ACCESS_TOKEN}}
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+> This is a modification package for debugging base on v0.5.2, please do not use. Please use the original package [vite-plugin-pwa](https://github.com/antfu/vite-plugin-pwa) .
+
 <p align='center'>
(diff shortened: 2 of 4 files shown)


antfu:
See d03cc13

chengpeiquan:
> See [d03cc13](https://github.com/antfu/vite-plugin-pwa/commit/d03cc138ee3f96278414087fcb4d0b891415c6f6)

Sorry to bother you again, I found that my understanding of the manifest was wrong last time. It seems that it must be read from the website directory, not from the CDN.

If use `https:/cdn.jsdelivr.net/gh/chengpeiquan/chengpeiquan.com@gh-pages/manifest.webmanifest`, it will load as `https://chengpeiquan.com/cdn.jsdelivr.net/gh/chengpeiquan/chengpeiquan.com@gh-pages/manifest.webmanifest`.

It start with `https://chengpeiquan.com/cdn.jsdelivr.net/` ……

If you have time, please look at it again. Need to adjust it to when the base use 'http....', in HTML file need to remove base as `/manifest.webmanifest`.

antfu:
I don't feel it's safe to make a special case for urls and make decisions for users. So I expose the option `base` for this plugin only so you can have better control over it.

PR #30, opened 2021-02-25
fix: add import script only when it's setted

lensanag:
Enable user disable the insertion of import script registerSW when 'injectRegister' is set disticnt from 'inline' or 'import'.
This make the plugin be unopinated of registerSW import tag on index.html.


lensanag:
Aditionally would it be add a configuration file for allow change the default name of register file (currently 'registerSW.js); This way the plugin will be agnostic of the filename set by the user.

PR #36, opened 2021-03-13
feat(networkfirst)

userquin:


userquin:
It would be useful to include suspense to test offline... At least to show while resolving child routes, it waits 10s on first network error

userquin:
@antfu can you review why suspense not working? has changed again with vue-router?

userquin:
It works only for first error, then the `Loading...` is missing: to reproduce it, stop server then F5 and then navigate to home.

PR #43, opened 2021-04-06
feat: inject manifest

### LINKED ISSUE #35: Resolve imports when bundling
As per [the workbox docs](https://developers.google.com/web/tools/workbox/guides/get-started#precaching), I have this in my service worker file:

```js
import { precacheAndRoute } from 'workbox-precaching'
precacheAndRoute(self.__WB_MANIFEST)
```

However, when I build my app, that part stays the same which results in the service worker not working because browsers can't execute js `import` statements in service workers (at least for now). It fails to execute the service worker with the error `Uncaught SyntaxError: Cannot use import statement outside a module`.

Not sure if this is an issue with the bundler or my understanding of workbox. All examples use import statements.

Relevant part of my `vite.config.js`:

```js
VitePWA({
	strategies: 'injectManifest',
	injectRegister: false, // service worker is registered from a separate script.
	injectManifest: {
		swSrc: './src/ServiceWorker/sw.js',
		swDest: './dist/sw.js',
	},
	manifest: {
		// manifest omitted
	}
}),
```

userquin:
closes #35

### DIFF
--- a/src/modules.ts
+++ b/src/modules.ts
@@ -1,8 +1,10 @@
-import { join, resolve } from 'path'
+import { join, resolve/*, relative */ } from 'path'
 import { promises as fs } from 'fs'
-import { execSync } from 'child_process'
+// import { execSync } from 'child_process'
 import { injectManifest } from 'workbox-build'
 // import { runEsbuild } from 'tsup'
+import { ResolvedConfig } from 'vite'
+import Rollup from 'rollup'
 import type { ResolvedVitePWAOptions } from './types'
 import { slash } from './utils'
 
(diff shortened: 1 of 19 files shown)


src/modules.ts
```diff
@@ -14,3 +17,63 @@ export async function generateRegisterSW(options: ResolvedVitePWAOptions, mode:
...
+  // this will not fail since there is an injectionPoint
+  options.injectManifest.swSrc = options.injectManifest.swDest
+  // options.injectManifest.mode won't work!!!
+  // error during build: ValidationError: "mode" is not allowed
+  if (options.injectManifest.mode)
+    delete options.injectManifest.mode
+
+  // inject the manifest
+  await injectManifest(options.injectManifest)
+  // after build the sw, we need to add process.env.NODE_ENV
```
userquin, line 71: We can remove this last step: we don't need to add `process.env.NODE_ENV`

userquin:
we also need to update `README.md`

userquin:
@antfu the example added is also  using the vuev irtual module and so it will work as vue-basic example, but instead defaults generated from workbox, we are now injecting the manifest: I just include the skip waiting section on `/public/sw.js` to make `Reload` button working.

userquin:
To run the example, just run `pnpm run build && npm run example:start:sw`

antfu (approved):
Looks solid, thank you!

PR #45, opened 2021-04-08
feat: allow typescript on service worker for injectManifest strategy

userquin:


userquin:
@antfu 

I left `srcDir` default to `public`  to avoid breaking changes (we cannot change it to `src` at this point...).

And review the README changes...

userquin:
One last thing, update all vite versions on examples: we are using vite 2.1.5 in the plugin but 3 examples are using vite 2.1.2.

I need to upgrade pnpm, if I do the upgrade I will revert `lockfileVersion: 5.3` to `lockfileVersion: 5.2` (pnpm lock file), tmr I'll upgrade pnpm.

PR #49, opened 2021-04-15
feat: add option to disable manifest generate, close #48

### LINKED ISSUE #48: Don't auto-create webmanifest
Hi!

Is it possible to NOT auto-create and inject the manifest.webmanifest but only generate the service worker?

My manifest needs to be loaded from a different server.

Thanks.

hannoeru:
close #48


stefanholzapfel:
wow you're quick :)

Issue #46, opened 2021-04-12
Inline workers break SW generation because they call closeBundle early


NotWoods:
When generating an inline web worker, Vite passes all its plugins to Rollup to create a bundle for the worker. Unfortunately that inner Rollup job [calls bundle.close and triggers closeBundle](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/plugins/worker.ts#L62) before any files are in the output folder. As a result, workbox-build throws an error because it can't find anything.

userquin:
@NotWoods can you include errors and vite and npm/pnpm/yarn versions? It seems to be a problem with `vite` itself.

userquin:
@NotWoods Can you provide a repro?

userquin:
The line linked is a bug that is being solved in vite,  the problem is that these line doesn't await promise returned and then is closed at any time later and in your case breaking the build.

userquin:
The line you link on vite should be like this: https://github.com/antfu/vite-plugin-pwa/blob/master/src/modules.ts#L52

In fact in your vite pr you have the same error, add await when closing the bundle...

userquin:
I write you on discord (`ViteLand`) on `plugins` chat: this must be fixed on `vite` side.

@antfu we should review if only including `vite:esbuild`, `vite:resolve` and `vite:terser` instead excluding, we need to do some tests.

userquin:
@NotWoods see https://github.com/antfu/vite-plugin-pwa/issues/68#issuecomment-859977520

PR #50, opened 2021-04-15
fix: update inject manifest build

userquin:
- include vite plugins instead of excluding them: this will prevent to break build when using other plugins on the target project (see #46 ).
- change output format to `es` instead of` cjs`.

### DIFF
--- a/src/modules.ts
+++ b/src/modules.ts
@@ -28,21 +28,25 @@ export async function generateInjectManifest(options: ResolvedVitePWAOptions, vi
   */
   // eslint-disable-next-line @typescript-eslint/no-var-requires
   const rollup = require('rollup') as typeof Rollup
-  // remove this plugin from the compilation: avoid infinite recursion
-  // remove also vite html transform and build to avoid rebuilding index.html
-  const excludedPluginNames = [
-    'vite-plugin-pwa',
-    'vite:build-html',
-    'vite:html',
+  const includedPluginNames = [
+    'alias',
+    'vite:dynamic-import-polyfill',
+    'vite:resolve',
+    'vite:esbuild',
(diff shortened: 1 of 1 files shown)


userquin:
@hannoeru the change of the first entry could also be included in `vite plugin worker`: just see `module.ts` changes.

hannoeru:
I think you also need to check service worker is esmodule or not then add `{ type: 'module' }` to registerSW & inlineRegister.

userquin:
Right now there is no browser supporting service worker modules.
This pr only only updates the way we pick rollup plugins from the target project and use `es` format instead `cjs` only to remove the `use strict`: the inject manifest Will include all dependencies from workbox dependencies ibto the sw bundle.
When some browser supports sw modules we can revoew this.

userquin:
@hannoeru it seems at least on chromium based browsers it will be available from 91 version: currently on version 90, see [Issue 824647: Implement ES Modules for service workers](https://bugs.chromium.org/p/chromium/issues/detail?id=824647).

About the `vite worker plugin` I mean: if a project including an inlined web worker also includes `vite-plugin-pwa` plugin, it will break project build, since `vite worker plugin` will include all plugins from the project, and so `vite-plugin-pwa` will be called when building the inlined `web worker`, but it is out of vite's build system order (it is called 4 or 5 steps before where it should be called). If in `vite worker.ts file` we just include only the plugins we need to build the inline worker, then problem solved (just filter rollup plugins similar to what we have in `src/modules.ts`).

hannoeru:
Got it 👍

userquin:
@hannoeru see issue 3010 on vite repo (I deal with infinite recursion here)

PR #59, opened 2021-05-11
fix: auto inject register if virtual register isn't requested

F-loat:
The resolved plugins will be in the following order:

> * Alias
> * User plugins with enforce: 'pre'
> * Vite core plugins
> * User plugins without enforce value
> * Vite build plugins
> * User plugins with enforce: 'post'
> * Vite post build plugins (minify, manifest, reporting)

The `generateBundle` plugin will be executed before `transformIndexHtml`  if `vite-plugin-pwa` and `transformIndexHtml` set `enforce: 'post'` both. And `options.injectRegister` will be wrong.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,7 +16,6 @@ export function VitePWA(userOptions: Partial<VitePWAOptions> = {}): Plugin[] {
   return [
     {
       name: 'vite-plugin-pwa',
-      enforce: 'post',
       apply: 'build',
       configResolved(config) {
         viteConfig = config


userquin:
@F-loat can you take a look at [chore: update inject manifest build](https://github.com/antfu/vite-plugin-pwa/pull/50#issuecomment-821098760) PR?

F-loat:
@userquin I'm not familiar with rollup, should I wait some plugins before injectServiceWorker. What is `rollup.input` for that.

userquin:
@F-loat this PR and linked seems to share a problem on vite's side.

The `rollup.input` entry is for apply `rollup build` the assets specified. In the PR we are building the `sw.js` from user input configured on `filename` option of `VitePWA` with `injectManifest` strategy, see `vue-basic-inject-manifest` example on `examples` directory.

F-loat:
Got it. Should we make it clear in the document, otherwise we are not very friendly to new people. In fact, currently you must specify `injectregister: 'script'` or `import virtual:pwa-register` to let it work.

PR #60, opened 2021-05-11
docs: update test steps

F-loat:
> [In order to facilitate local development, localhost is considered a secure origin by browsers as well. ](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers#setting_up_to_play_with_service_workers)

So `vite preview` worked well.


userquin:
@F-loat does lighthouse test works without https? Can you run it on chrome?

F-loat:
The lighthouse works, too.
image

Issue #55, opened 2021-04-28
How disable `injectManifest` ?


cawa-93:
I Try use this plugin in Electron app. 

Technically, this is a web page with a strict Content-Security-Policy.
During development, the page runs at `http://localhost:3000`, and in the production environment as `file:///C:/path-to/app/resources/index.html`

Precache, I think, is not very important for my, because all assets are located locally anyway. But using the schema `file:` imposes certain problems for the service worker.

Therefore, I am only interested in the service worker as a tool for caching network requests. And I want to completely disable `injectManifest`.

I tryed
```js
    VitePWA({
      injectRegister:'script',
      manifest: false,
      injectManifest: false,
})
```
But in `sw.js` still has 
```js
  workbox.precacheAndRoute([{
    "url": "index.8daf1fe7.css",
    "revision": "bb76d49a4d80950ac285a814832c9c8a"
  }, {
    "url": "index.f66a5a19.js",
    "revision": "a8e69220ac0a4822a5e98ec359765139"
  }, {
    "url": "index.html",
    "revision": "7c2b1f2f6f2698a7dd5ba69c1e4ca1c3"
  }, {
    "url": "registerSW.js",
    "revision": "e21bbf02f2e95e944e8acae5f65ca7f7"
  }, {
    "url": "vendor.f57ce527.js",
    "revision": "cbcca323ae9391f4ef2f13c0065e7abd"
  }, {
    "url": "Watch.8c7a9f24.js",
    "revision": "b7aceb22f06e7a44553c43001c7c1f60"
  }], {});
  ```

**Comment by userquin** [comment]
@cawa-93 try this: 
```javascript
VitePWA({ injectRegister: null })
```

cawa-93:
@userquin It not help. Generated `sw.js`:

```js
  workbox.precacheAndRoute([{
    "url": "index.8daf1fe7.css",
    "revision": "bb76d49a4d80950ac285a814832c9c8a"
  }, {
    "url": "index.f66a5a19.js",
    "revision": "a8e69220ac0a4822a5e98ec359765139"
  }, {
    "url": "index.html",
    "revision": "dd0936008c2fede9714e474109432223"
  }, {
    "url": "vendor.f57ce527.js",
    "revision": "cbcca323ae9391f4ef2f13c0065e7abd"
  }, {
    "url": "Watch.8c7a9f24.js",
    "revision": "b7aceb22f06e7a44553c43001c7c1f60"
  }], {});
[...]

PR #62, opened 2021-05-15
 feat: static assets handling

### LINKED ISSUE #31: Automatically add manifest icons to build dir
Hi, I don't know if it's a bug or a feature request but it would be fantastic if the images listed in the `icons[]` array of the manifest would be automatically added to the build dir (`dist/assets` by default).

### LINKED ISSUE #53: include custom files/dirs in sw cache?
Hello,
I have a directory `/public/images` which contain icons. the icons are referenced in `index.html` and the pwa webmanifest as `/images/icon-x.png`.
The images directory is moved to the dist folder on build, as expect. but it is not included in the cache, which is kinda expected as well.
Can I include it manually? The images disappear right after reload when the app goes offline.

Note: I've tried using the `assetsInclude` config of Vite, as the PWA `manifest` config don't automatically change URLs to generated ones, meaning in webmanifest `/images/icon-x.png` don't turn into /images/icon-x-123ea.png` on build.

demo link if needed: https://vite-alpine-pwa.vercel.app/

### LINKED ISSUE #54: Pictures cannot be accessed when offline
Modify based on this example
[https://github.com/antfu/vite-plugin-pwa/tree/master/examples/vue-router](https://github.com/antfu/vite-plugin-pwa/tree/master/examples/vue-router)

vite.config.js
```
 VitePWA({
      registerType: "autoUpdate",
      manifest: {
        version: "1.1",
        comment: "some comment",
        lang: "cn",
        name: "test",
        scope: "/",
        display: "fullscreen",
        start_url: ".",
        short_name: "test",
        description: "test66",
        orientation: "portrait",
        background_color: "#ffffff",
        theme_color: "#ffffff",
        generated: "true",
        icons: [
          {
            src: "/public/assets/images/icon.png",
            sizes: "144x144",
            type: "image/png",
          },
        ],
      },
    })
```
packjson.js
```
"dependencies": {
    "axios": "^0.21.1",
    "vue": "^3.0.11",
    "vue-router": "^4.0.6"
  },
  "devDependencies": {
    "@vitejs/plugin-vue": "^1.2.1",
    "@vue/compiler-sfc": "^3.0.11",
    "animejs": "^3.2.1",
    "eslint": "^7.24.0",
    "eslint-plugin-vue": "^7.9.0",
    "sass": "^1.32.10",
    "vite": "^2.2.1",
    "vite-plugin-pwa": "^0.7.2",
    "vite-plugin-svg-icons": "^0.4.3",
    "vite-plugin-windicss": "^0.14.6",
    "windicss": "^2.5.14"
  }
```

In /src/pages/About   
```
<template>
  <div>About route, built at: {{ date }} ({{ timeAgo }})</div>
  <br />
  <button @click="router.push('/')">Home</button>
  <img src="/src/img/logo.png" alt="" />
</template>
```
`yarn start`
open https://localhost  Logo no problem

Refresh or re access the address after offline, and the logo will not be displayed
Want to realize all data cache, how to operate
There may be a lot of pictures in the future

userquin:
closes #53 
closes #54 
closes #31


userquin:
We need to review a few things on `src/options.ts`. For exmaple, this PR will handle assets in `publicDir`, but I'm not sure if using Vite `publicDir` option or plugin manifest  `publicDir` option.

I have added:
- `Static assets handling` on README.md
- modified the vue-router sample including some icons and including the svg favicon on App.vue.

antfu (approved):

antfu:
Should use Vite's `publicDir` I think.

@userquin since you are pretty active on this project, do yo want to take the lead and managing the release process as well? Thank a lot

userquin:
@antfu Thank you very much Anthony, but the creature is yours. I have no problem keeping an eye on new issues and including new features via PR, and don't worry if I have to wait for you to review them ;) and release it.

userquin:
@antfu the PR is using Vite's `publicDir` and so it is ready: all static assets on this directory will be added, via PWA manifest icons or via `include` option. 

By default all icons listed on PWA manifest icons will be included, and so all projects updating to the new release will be updated (for example `Vitesse` and `@vueuse`).

I'll try to see if I can make a new PR to automatically include `favicon`, I need to find a way to detect them on `index.html`.

userquin:
@antfu We also need to add `manifest.webmanifest` to the service worker *precache*. I'm working on it and since it is build by the plugin I need to find the right place to include it, since it will be not generated/present and its `revision` cannot be calculated until generated.

userquin:
We also need to add the enties when using `inject manifest` strategy. I'll update this PR with and update the example with vue-router changes.

userquin:
Finally, we need to handle Vite's `base`. I'll exclude if using `Full url base`, the service worker will no include external resources. I'm working on it...

userquin:
@antfu this PR is ready to review again. I've tested that works on `vue-router` and `vue-basic-inject-manifest`, the static assets are configured for both.

We need to do a test with a base <> '/', seems that `https-localhost` cannot configure a context and breaks some testing I do (`/test/`): it returns `manifest.webmanifest` content with `index.html` content.

userquin:
I need to fix resolving icons paths: for exmaple if using `/someicon.png` we need to convert it to relative. I'm fixing it...

src/assets.ts
```diff
@@ -0,0 +1,148 @@
...
+  } = resolvedVitePWAOptions
+
+  // static assets handling
+  // we need to check inject manisfest strategy
+  // additionalManifestEntries will go to workbox entry
+  // or to injectManifest entry
+  const { publicDir } = viteConfig
+  const useInjectManifest = strategies === 'injectManifest'
+  // include static assets
+  const includeUrl: string[] = []
```
userquin, line 92: modify `include` type adding `false`, defaults to undefined so we need to check if `undefined` (I'll review defaults to also include some apple icons, see picture below):

`['favicon.ico', 'favicon-<size>.png', 'favicon.svg', 'safari-pinned-tab.svg'>]`

where `<sizes>` will be: `16x16`,`32x32`, `48x48`, `72x72`, `96x96`

then merge with provided on `include` option.

We also need to add this to `README.md` file.

imagen

userquin: If we include these defaults, then update `vue-router` and `inject-manifest` examples removingth `favicon.svg` from `include` option.

userquin: maybe it will be better to do something similar to Vite `assets` plugin (see [assetsInclude configuration option](https://vitejs.dev/config/#assetsinclude)): `include: string | Regex | (string | Regex)[] | undefined` and force the user to include all static assets.

README.md
```diff
@@ -147,6 +148,16 @@ To resolve service worker types, just add `WebWorker` to lib entry on your `tsco
 "lib": ["ESNext", "DOM", "WebWorker"],
 ```
 
+### Static assets handling
+
+By default, all icons on `PWA Manifest` option found under Vite's `publicDir` option directory, will be included 
+in the service worker *precache*. You can disable this option using `includeManifestIcons: false`.
+
+You can also add another static assets such as `favicon`, `svg` and `fonts` files using `include` option.
+You don't need to configure `PWA Manifest icons` on `include` option.
```
userquin, line 157: Add entry specifiying that static assets must be under Vite's `publicDir` option directory.

userquin:
We can use Vite's assets plugin helper functions like [checkPublicFile](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/plugins/asset.ts#L120) to resolve static assets here.

src/assets.ts
```diff
@@ -0,0 +1,148 @@
...
+        addManifestEntry(
+          publicDir,
+          icon.src as string,
+          includeUrl,
+          additionalManifestEntries,
+        )
+      })
+    }
+    // manifest.webmanifest
+    addWebManifestEntry(
```
userquin, line 138: We need to check if `manifest.webmanifest` exists on `publicDir`, and change the function call to calculate the hash from file.

Maybe we need also load icons from existing `manifest.webmanifest`.

userquin: forgot this

userquin:
We need to review this line on `index.ts`:

`if (options.injectRegister === 'script' && !existsSync(join(viteConfig.root, 'public', FILE_SW_REGISTER))) {`

Since we have all information, we must check using Vite's `publicDir`, here we are checking `public` directory and can be wrong.

src/assets.ts
antfu: Could simply accepts an array of globs, and pass it to `fast-glob` to get the file list. Also for auto included icons, we could run a unique so ppl don't need to care if they should exclude them or not.

antfu:
> Since we have all information, we must check using Vite's publicDir, here we are checking public directory and can be wrong.

That's true. At the time this plugin with written (vite 1.x) there isn't a way to configure it IIRC. You could either do that in this PR or another one. Thanks!

src/types.ts
```diff
@@ -85,6 +85,20 @@ export interface VitePWAOptions {
...
    */
   base?: string
+  /**
+   * `public` resources to be added to the PWA manifest.
+   *
+   * You don't need to add `manifest` icons here, it will be auto included.
+   *
+   * The `public` directory will be resolved from Vite's `publicDir` option directory.
+   */
+  include: string | string[] | undefined
```
antfu, line 95:
```suggestion
  includeAssets: string | string[] | undefined
```

Maybe this way could be more explicit, WDYT?

userquin: yeah, I forgot to mention renaming options...

userquin: rename done

src/assets.ts
userquin: But, how to resolve the url path added to the `__SW_MAINIFEST`? For manifest icons we can assume to be under Vite's `publicDir` option directory, but for `includeAssets`?

antfu: yeah, both of them should be under the publicDir. Or do you see any other cases?

userquin: ok, so we only need to check if under Vite's  `publucDir` once globs resolved?

userquin: OK, we also need to change `configResolved` to be `async` on `index.ts`.

antfu: If it's hard, we can use `fg.sync` for sync. But would prefer async for performance if possible.

userquin: no problem, just adding `async` to `configResolved`and `resolveOptions` and done

userquin: imagen

userquin: can I use `Promise.allSettled` to do md5 computation parallel or just sequential with `Promise.all`?

userquin:
I forgot this entry: https://github.com/antfu/vite-plugin-pwa/issues/55#issuecomment-828608818, we can rewrite `assets` to use `globPatterns` collecting it from the logic...

NickClark:
Needing this as well. Is there a workaround in the meantime? Being blocked by not having icons and also images not working when offline.

userquin:
@NickClark yes, there is a workaround, see https://github.com/antfu/vite-plugin-pwa/issues/53#issuecomment-841672553 and comments after it.

You can see also last paragraph about `globPatterns`: [static-assets-handling](https://github.com/antfu/vite-plugin-pwa/blob/d3f6b6821aa8bd8ef264e0afd5ead4f5d9933f10/README.md#static-assets-handling) and link [here](https://github.com/antfu/vite-plugin-pwa/issues/55#issuecomment-828602464)

PR #69, opened 2021-05-29
chore: add more options to ManifestOptions

rehhouari:
Added some more missing types for ManifestOptions and and a catch-all at the end in case some new ones are added that this library don't document

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -147,4 +147,52 @@ export interface ManifestOptions {
    * @default A combination of `routerBase` and `options.build.publicPath`
    */
   publicPath: string
+  /**
+   * @default []
+   */
+  related_applications: {
+    platform: string,
+    url: string,
+    id?: string,
+  }[]
+  /**
+   * @default false
+   */
+  prefer_related_applications: boolean
+  /**
+   * @default []
+   */
+  protocol_handlers: {
+    protocol: string,
+    url: string,
+  }[]
+  /**
+   * @default []
+   */
+  shortcuts: {
+    name: string,
+    short_name?: string,
+    url: string,
+    description?: string,
+    icons: []string,
+  }[]
+  /**
+   * @default []
+   */
+  screenshots: {
+    src: string,
+    sizes: string,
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -147,4 +147,52 @@ export interface ManifestOptions {
...
+   */
+  related_applications: {
+    platform: string,
+    url: string,
+    id?: string,
+  }[]
+  /**
+   * @default false
+   */
+  prefer_related_applications: boolean
```
userquin, line 161: should be `prefer_related_applications?: boolean` ([spec](https://developer.mozilla.org/en-US/docs/Web/Manifest/prefer_related_applications))

src/types.ts
```diff
@@ -147,4 +147,52 @@ export interface ManifestOptions {
    * @default A combination of `routerBase` and `options.build.publicPath`
    */
   publicPath: string
+  /**
+   * @default []
+   */
+  related_applications: {
```
userquin, line 153: should be `related_applications?: {...}[]` [spec](https://developer.mozilla.org/en-US/docs/Web/Manifest/prefer_related_applications)

userquin:
@rehhouari please review mandatory entries, all entries you added are no mandatory.

The `catch-all` entry should not be added, since it can be anything in the future, I'll prefer add them when necessary and have them typed, since these entries shouldn't change along time.

userquin:
@rehhouari or add defaults to `options.ts` module

imagen

src/types.ts
rehhouari: yes true sorry!

rehhouari:
Like this?

userquin:
@rehhouari just remove `catch-all`, keep original `options.ts`, there is no need to add those defaults, and sorry for the inconveniences. I forgot we are using `Partial` and so there is no need to include defaults for those guys.

And thx for the PR.

rehhouari:
done ^^

userquin (approved):

PR #79, opened 2021-06-13
feat: network first strategy

userquin:
Initial version for network first strategy (fully functional):
- can provide credentials and timeout (only for `Network First Cache Strategy` not for `Custom Cache Network Race Strategy`).
- ~~by default using [Custom Cache Network Race Strategy](https://developers.google.com/web/tools/workbox/modules/workbox-strategies#custom_cache_network_race_strategy)~~.
- [Custom Cache Network Race Strategy](https://developers.google.com/web/tools/workbox/modules/workbox-strategies#custom_cache_network_race_strategy)
- by default, the cache strategy is `Network First Cache Strategy` since `Custom Cache Network Race Strategy` is for special cases (when using small chunks)

To run examples, from root directory, first run `pnpm install && pnpm run build` and then:
-  [Custom Cache Network Race Strategy](https://developers.google.com/web/tools/workbox/modules/workbox-strategies#custom_cache_network_race_strategy): `pnpm run example:start:networkfirst:race`.
- `Network First Cache Strategy`: `pnpm run example:start:networkfirst:no-race`.

I was forced to add `skip waiting` logic, just because the service worker once updated needs to be activated. If we omit firing `skip event` from service worker, then the cache will no be cleared (at least `index.js` asset).

Do not merge yet, I need to review it for development, it is not yet protected.


userquin:
@MartinMalinda

userquin:
@antfu branch `sw-networkfirst` can be removed.

userquin:
The name of the directory `src/client/workbox-recipes` makes sense to us, but seen from the outside it can give you an understanding that it is really a set of `workbox recipes`, and can be used as `workbox` entry explains: just see the intro from [Workbox Recipes site] (https: / /developers.google.com/web/tools/workbox/modules/workbox-recipes) .

That's not true, the `network-first.ts` module is an `esm module` we use to just replace the placeholders, and so maybe we need to rename that directory.

userquin:
About the skip waiting problem, you can just comment out this at end of:
```ts
// src/client/workbox-recipes/network-first.ts
...
/*
// this is necessary, since the new service worker will keep on skipWaiting state
// and then, caches will not be cleared since it is not activated
self.addEventListener('message', async(event) => {
  if (event.data && event.data.type === 'SKIP_WAITING')
    await self.skipWaiting()
})
*/
```

then `pnpm run build` and rebuild the `race` and `no-race` examples 2 times (first one to install it, `offline ready dialog` and last one to see the service worker is waiting) and check the new service worker on the 2 installation keeps on `skip waiting state` in yellow on `chrome dev tools > Application > Service Workers`.

 If you see the console on `chrome dev tools` there will not be there checking outputs.

If you then, add again the `skip waiting event` on `src/client/workbox-recipes/network-first.ts`, and build `race` and `no-race` again, you will see the `service worker` will be activated and on console you will see the entries removed from the cache.

userquin:
I'm testing with `autoUpdate` logic, and so we can simplify `register.ts` module logic:
```ts
clientsClaim()
self.skipWaiting()
// this is necessary, since the new service worker will keep on skipWaiting state
// and then, caches will not be cleared since it is not activated
// self.addEventListener('message', async(event) => {
//   if (event.data && event.data.type === 'SKIP_WAITING')
//     await self.skipWaiting()
// })
```
and so, `register.ts` only need small changes from the original.

userquin:
Also tested that last changes works with other `strategies` and `registerType` options.

userquin:
@antfu 

About review for development, the virtual module is the dummy one, and since it is not for build we are done, and so, this PR is ready to review.

src/types.ts
```diff
@@ -99,6 +99,29 @@ export interface VitePWAOptions {
    * option directory.
    */
   includeManifestIcons: true
+  /**
+   * Configuration for `networkFirst` strategy.
+   * Keep it here if in future we need to extend its configuration.
+   */
+  networkFirst?: {
+    /**
+     * Should we use `Custom Cache Network Race Strategy` or `NetworkFirst`?
```
userquin, line 108: This should be:
```md
Should we use `Custom Cache Network Race Strategy` or `Network First Cache Strategy`?
```

src/types.ts
```diff
@@ -99,6 +99,29 @@ export interface VitePWAOptions {
...
+    /**
+     * @default false
+     */
+    debug?: boolean
+    /**
+     * @default 'same-origin'
+     */
+    credentials?: RequestCredentials
+    /**
+     * Only for `NetworkFirst`.
```
userquin, line 121: Should be:
```md
Only for `Network First Cache Strategy`.
```

README.md
```diff
@@ -195,6 +196,74 @@ To resolve service worker types, just add `WebWorker` to lib entry on your `tsco
...
+`injectManifest`. 
+
+There are 2 available cache strategies:
+- [Custom Cache Network Race Strategy](https://developers.google.com/web/tools/workbox/modules/workbox-strategies#custom_cache_network_race_strategy)
+- `Network First Cache Strategy`
+
+By default, the service worker will use `Custom Cache Network Race Strategy`.
+You can see an explanation [here](https://jakearchibald.com/2014/offline-cookbook/#cache--network-race).
+
+To configure `Network First Cache Strategy` instead `Custom Cache Network Race Strategy`:
```
userquin, line 211: This entite block, should be moved to configuration entry

src/types.ts
```diff
@@ -99,6 +99,29 @@ export interface VitePWAOptions {
...
+  /**
+   * Configuration for `networkFirst` strategy.
+   * Keep it here if in future we need to extend its configuration.
+   */
+  networkFirst?: {
+    /**
+     * Should we use `Custom Cache Network Race Strategy` or `Network First Cache Strategy`?
+     * @default true
+     */
+    raceStrategy?: boolean
```
userquin, line 111: I think we should change to `Network First Cache Strategy` by default

userquin: Since `Custom Cache Network Race Strategy` is for special case (when using smalls chunks) I'll keep there but setting  `Network First Cache Strategy` as default.

I'll change the `jsdoc` on `types.ts`, the `README.md` and modify the logic a little bit.

userquin:
@MatinMalinda @antfu I close this PR since the code provided on `src/client/workbox-recipes/network-first.ts` module can be used with this plugin to generate a custom service worker. 

I'll include an example in the new docs.

FezVrasta:
@userquin may you point to the network-first example? I can't find anything in the docs.

userquin:
@FezVrasta https://github.com/userquin/vite-plugin-pwa/tree/feat/network-first-workbox-recipe

It should be on the top of this PR, the changes can be found on `Files changed` tab.

PR #94, opened 2021-07-27
feat: add service worker registration callbacks

### LINKED ISSUE #93: How about add `onRegitered` event  to `registerSW` ?
I'm trying to trigger updates manually like google's document says: https://developers.google.com/web/fundamentals/primers/service-workers/lifecycle#manual_updates

At the moment, we can that using [`navigator.serviceWorker.getRegistration`](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/getRegistration) .

```ts
const updateSW = registerSW({
  onNeedRefresh: () => {
    setOpenDialog(true);
  },
});
navigator.serviceWorker.getRegistration('/').then((reg) => {
  // do something
  setInterval(() => {
    reg?.update();
  }, INTERVAL);
});
```

But, if there are `onRegistered` event that has `ServiceWorkerRegistration` argument, be able to become the above code simpler (and safe(?)).

```ts
const updateSW = registerSW({
  onNeedRefresh: () => {
    setOpenDialog(true);
  },
  onRegistered: (reg) => {
    // do something
    setInterval(() => {
      reg?.update();
    }, INTERVAL);
  }
});
```

How about this?

userquin:
This PR includes:
- add sw register callback
- add sw register error callback
- ~~add an example for logging to console the registration on router example (`ReloadPromp.vue`)~~
- add a working example on `vue-router` (configurable on `ReloadPrompt.vue` via `@rollup/plugin-replace`)

closes #93

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,6 +225,32 @@ VitePWA({
 })
 ```
 
+### Configure application to check for Service Worker updates
+
+As explained in `Manual Updates` entry on `The Service Worker Lifecycle` [here](https://developers.google.com/web/fundamentals/primers/service-workers/lifecycle#manual_updates),
+you can use this code to configure this behavior on your application:
+
+```ts
+// main.ts
+import { registerSW } from 'virtual:pwa-register'
+
+const updateSW = registerSW({
+  // other options
+  onRegisted(r) {
+    r && setInterval(() => {
+      r.update()
+    }, 60 * 60 * 1000 /* 1 hour: timeout in milliseconds */)
+  }
+})
+```
+
+**Warning**: since `workbox-window` uses a time-based `heuristic` algorithm to handle service worker updates, if you 
+build your service worker and register it again, if the time between last registration and the new one is less than 
+1 minute, then, `workbox-window` will handle the `service worker update found` event as an external event, and so the
+behavior could be strange (for example, if using `prompt`, instead showing the dialog for new content available, the 
+ready  to work offline dialog will be shown; if using `autoUpdate`, the ready to work offline dialog will be shown and 
+shouldn't be shown).
+
 ### Full config
 
 Check out the type declaration [src/types.ts](./src/types.ts) and the following links for more details.
--- a/README.md
+++ b/README.md
@@ -236,7 +236,7 @@ import { registerSW } from 'virtual:pwa-register'
 
 const updateSW = registerSW({
   // other options
-  onRegisted(r) {
+  onRegistered(r) {
     r && setInterval(() => {
       r.update()
     }, 60 * 60 * 1000 /* 1 hour: timeout in milliseconds */)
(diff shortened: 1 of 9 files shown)


antfu (approved):

userquin:
Don't merge yet, I need to check if the `registration` is safe once new update found. I also need to add a test, maybe on `router` and also on `inject manifest` examples.

userquin:
I need to check the logic we need to update the sw, it seems that update works but the sw keeps on skip waiting state, so the sw is not yet updated. I need to find the way to send skip waiting event to the sw and so the prompt the user or auto update:

imagen

imagen

userquin:
Well, it seems to work, I'm finishing the examples, but:

Sometimes the ready to work dialog is shown instead the new update found dialog. I think the problem is here (`node_modules/.pnpm/workbox-window@6.1.5/node_modules/workbox-window/src/Workbox.ts`):

imagen

This check is problematic, the last check (see bellow how `updateLikelyTriggeredExternally` is calculated): if we build  and register the SW in less than 1 minute (from previous registration), then,  it will be handled as external:

```ts
 // Workbox.ts: line 341
 private readonly _onUpdateFound = () => {
    // `this._registration` will never be `undefined` after an update is found.
    const registration = this._registration!;
    const installingSW = registration.installing as ServiceWorker;

    // If the script URL passed to `navigator.serviceWorker.register()` is
    // different from the current controlling SW's script URL, we know any
    // successful registration calls will trigger an `updatefound` event.
    // But if the registered script URL is the same as the current controlling
    // SW's script URL, we'll only get an `updatefound` event if the file
    // changed since it was last registered. This can be a problem if the user
    // opens up the same page in a different tab, and that page registers
    // a SW that triggers an update. It's a problem because this page has no
    // good way of knowing whether the `updatefound` event came from the SW
    // script it registered or from a registration attempt made by a newer
    // version of the page running in another tab.
    // To minimize the possibility of a false positive, we use the logic here:
    const updateLikelyTriggeredExternally =
        // Since we enforce only calling `register()` once, and since we don't
        // add the `updatefound` event listener until the `register()` call, if
        // `_updateFoundCount` is > 0 then it means this method has already
        // been called, thus this SW must be external
        this._updateFoundCount > 0 ||
        // If the script URL of the installing SW is different from this
        // instance's script URL, we know it's definitely not from our
        // registration.
        !urlsMatch(installingSW.scriptURL, this._scriptURL) ||
        // If all of the above are false, then we use a time-based heuristic:
        // Any `updatefound` event that occurs long after our registration is
        // assumed to be external.
        (performance.now() >
            this._registrationTime + REGISTRATION_TIMEOUT_DURATION) ?
                // If any of the above are not true, we assume the update was
                // triggered by this instance.
                true : false;
```

userquin:
I need to add some docs, don't merge yet.

userquin:
I have added 2 new scripts to run `vue-router` example with the `check for update logic`, just use (run `pnpm run build` before):
- `pnpm run example:router:start:reloadsw` to run the example with `prompt`.
- `pnpm run example:router:start:claims:reloadsw` to run the example with `autoUpdate`.

antfu (approved):

PR #115, opened 2021-08-19
docs: fix seo mobile anchors and nav links for mobile

userquin:
This PR includes:
- changed all markdown lists items using new `MdListAnchor.vue` wrapped with `ul` to increase clickeable area: also resolves seo mobile anchors issues.
- updated all media queries to use 780px instead 720px to avoid nav links overlap with the nav title when changing size
- fix for nav links when included on left menu: right now, entries cannot be expanded with mouse: missing `:class="{ open }"` on `NavDropdownLink.vue`.

### DIFF
--- a/docs/.vitepress/theme/components/MdListAnchor.vue
+++ b/docs/.vitepress/theme/components/MdListAnchor.vue
@@ -1,0 +1,68 @@
+<script setup lang="ts">
+import { useRouter } from 'vitepress'
+
+const props = defineProps<{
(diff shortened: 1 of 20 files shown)


userquin:
@antfu before merge change `Typescript` with `TypeScript`, just search and replace in md files: for example, here https://vite-plugin-pwa.netlify.app/guide/inject-manifest.html#typescript-support and the first paragraph.

PR #126, opened 2021-08-24
feat: use vite `build.sourcemap` to configure `workbox sourcemap`

userquin:
When using `generateWS` strategy, this PR configures `workbox.sourcemap` with `build.sourcemap` vite option:
- if `workbox.sourcemap` is configured, its value is not changed
- otherwise, if `build.sourcemap` vite option is `true`, `inline` or `hidden` then `workbox.sourcemap` is set to `true` otherwise to `false` 

**NOTE**: the default `build.sourcemap` vite option value is `false`, while the default `workbox.sourcemap` option value is `true`. With this PR, we have aligned both.

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -101,6 +101,12 @@ export async function resolveOptions(options: Partial<VitePWAOptions>, viteConfi
     workbox.clientsClaim = true
   }
 
+  // use vite build.sourcemap to configure `generateSW` sourcemap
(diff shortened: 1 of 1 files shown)


userquin:
@antfu Once merged (and released: 0.11.1), I'll add a hint on `workbox` entry on docs...

antfu (approved):

PR #128, opened 2021-08-28
docs: add click outside on side bar menu for mobile

userquin:
This PR includes:
- Hides the side bar menu on mobile when click outside: added `@vueuse/shared` to use `useDebounceFn`
- fix strategy with `generateSW` instead `prompt` on `GenerateSWCleanupOutdatedCaches` component

@antfu Just review it, I'll merge (all logic on `Layout` component).

### DIFF
--- a/docs/.vitepress/theme/Layout.vue
+++ b/docs/.vitepress/theme/Layout.vue
@@ -3,8 +3,10 @@ import {
   useRoute,
   useData,
 } from 'vitepress'
-import { onClickOutside } from '@vueuse/core'
-import { useDebounceFn } from '@vueuse/shared'
+import {
+  onClickOutside,
+  useDebounceFn,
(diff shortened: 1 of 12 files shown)


userquin:
Tested preview that works at least on android with chrome

docs/.vitepress/theme/Layout.vue
```diff
@@ -3,6 +3,8 @@ import {
   useRoute,
   useData,
 } from 'vitepress'
+import { onClickOutside } from '@vueuse/core'
+import { useDebounceFn } from '@vueuse/shared'
```
antfu, line 7: Just use '@vueuse/core' then

userquin: updated

antfu (approved):

PR #159, opened 2021-11-17
docs(solid-js): promote Show component usage rather than ternary

davedbase:
The SolidJS core team likes to promote the use of `<Show>` component with the fallback attribute as a more expressive way of displaying elements. I made a small revision to this example. Thanks for including SolidJS in the examples, it's quite helpful.

BTW we're using vite-plugin-pwa on solidjs.com =)

### DIFF
--- a/docs/frameworks/solidjs.md
+++ b/docs/frameworks/solidjs.md
@@ -41,7 +41,7 @@ You can use this `ReloadPrompt.tsx` component:
 
 ```tsx
 // eslint-disable-next-line no-use-before-define
-import { Component } from "solid-js";
+import { Component, Show } from "solid-js";
 import styles from './ReloadPrompt.module.css'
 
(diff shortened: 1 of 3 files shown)


userquin:
@davedbase can you also modify the example on `packages/solid-router`?

userquin:
If the PR comes from editing on pwa docs, I'll update the example.

davedbase:
@userquin I assume you meant examples/solid-router. I went ahead and made the adjustment. =)

userquin:
@davedbase yeah, sorry, too many repos ;)

userquin:
@davedbase on `docs/frameworks/solidjs.md` add the import for `Show` component:
```ts
import { Component } from "solid-js";
```

should be

```ts
import { Component, Show } from "solid-js";
```

davedbase:
Good catch @userquin. Fixed.

PR #173, opened 2021-11-28
fix(react): `registerSW` inside `useState`

### LINKED ISSUE #171: react: `onRegistered` is called a lot
I'm using `useRegisterSW` from `virtual:pwa-register/react` and the `onRegistered` callback is called more than once on a given page load.

I would expect this would only be called once per page load. I looked at the `useRegisterSW` hook and it appears to register the SW outside of a `useEffect`. It seems like this should be inside a `useEffect(fn, [])` call so it only gets called on component mount, not every render.

image

Generally it appears 3x when I'm loading my page, but I've had it sometimes appear a 4th time. I'm using `onRegistered` to start an interval to check for SW updates, and I don't want 4 intervals running at once.

userquin:
It seems service worker registration is being called multiple times preventing workbox window to work correctly.

@antfu Don't release yet, I'm fixing also `preact`, from github I don't know how to modify 2 files in the same PR.

closes #171

### DIFF
--- a/src/client/build/react.ts
+++ b/src/client/build/react.ts
@@ -16,18 +16,20 @@ export function useRegisterSW(options: RegisterSWOptions = {}) {
   const [needRefresh, setNeedRefresh] = useState(false)
   const [offlineReady, setOfflineReady] = useState(false)
 
-  const updateServiceWorker = registerSW({
-    immediate,
-    onOfflineReady() {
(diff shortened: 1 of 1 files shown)


userquin:
`preact` also fixed: both examples tested on local via `pnpm run examples` and works as expected

PR #182, opened 2021-11-30
feat: support useCredentials option to enable crossorigin attribute

### LINKED ISSUE #178: Make crossorigin attribute for link[rel="manifest"] configurable
When a site sits behind authentication, you need to use

```html
<link rel="manifest" href="/..." crossorigin="use-credentials" />
```

so that the browser sends credentials along with the request and thus receives the manifest instead of a 401.

jonaskuske:
fix #178 

Verified it works in the vue example as requested in https://github.com/antfu/vite-plugin-pwa/issues/178#issuecomment-982097744.
Not sure if I was supposed to leave the change to the PWA config (`useCredentials: true`) in `examples/vue-router/vite.config.ts` so I didn't commit it for now, but that's a quick change.

### DIFF
--- a/docs/guide/faq.md
+++ b/docs/guide/faq.md
@@ -19,6 +19,16 @@ add the following to the `compilerOptions.types` array of your `tsconfig.json`:
 }
 ```
 
+## Web app manifest and 401 status code (Unauthorized)
+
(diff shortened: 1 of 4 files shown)


userquin (approved):
there is no need to add `useCredentials` on the examples.

the `html.ts` module can be optimized but since it is called only once on build, keep as it is rn

PR #185, opened 2021-12-01
feat: add `id` to manifest type

nstringham:
adds `id` to ManifestOptions so that I can set an id without getting a typescript error

https://www.w3.org/TR/appmanifest/#id-member

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -140,7 +140,7 @@ export interface ManifestOptions {
    */
   scope: string
   /**
-   * a string that represents the identity for the application
+   * A string that represents the identity for the application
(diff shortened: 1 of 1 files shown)


src/types.ts
```diff
@@ -139,6 +139,10 @@ export interface ManifestOptions {
    * Restricts what web pages can be viewed while the manifest is applied
    */
   scope: string
+  /**
+   * a string that represents the identity for the application
```
userquin, line 143: replace with: A string that represents the identity for the application

userquin (approved):

PR #220, opened 2022-01-29
fix: shortcuts icons type on pwa manifest

### LINKED ISSUE #216: Incorrect type for icons property in manifest
Previously, I have opened an issue at #213, but even after following the solutions, the error did not resolve.

The type definitions for "icons" in the manifest might be incorrect.
I am receiving the following type warning:
image
image

I am a newbie at this, but this might be coming from the type definitions file (/dist/index.d.ts), the type of icons are mentioned as icons: Record<string, any>[];

userquin:
https://w3c.github.io/manifest/#processing-image-resources

breaking change?

closes #216

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -210,7 +210,7 @@ export interface ManifestOptions {
     short_name?: string
     url: string
     description?: string
-    icons: string[]
+    icons?: Record<string, any>[]
   }[]
   /**
    * @default []
--- a/src/types.ts
+++ b/src/types.ts
@@ -210,7 +210,7 @@ export interface ManifestOptions {
     short_name?: string
     url: string
     description?: string
-    icons?: Record<string, any>[]
+    icons: Record<string, any>[]
   }[]
   /**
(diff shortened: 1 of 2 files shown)


userquin:
adding also shortcuts icons to the sw precache manifest

src/types.ts
```diff
@@ -210,7 +210,7 @@ export interface ManifestOptions {
     short_name?: string
     url: string
     description?: string
-    icons: string[]
+    icons: Record<string, any>[]
```
antfu, line 213: Should support both type and do a normalization in runtime

userquin: from 1 string to this https://www.w3.org/TR/image-resource/#processing-an-image-resource-from-json? we need to add some defaults for `sizes` (`96x96`) and `type` (`image/png`) and we can omit `purpose`: what about more than 1 icon? 

Maybe it would be better adding the string type union and throw error or show warning when using string.

PR #244, opened 2022-03-11
perf: lazy load workbox-build to reduce config loading time

sapphi-red:
I was thinking that vite's starting up time was slow in [my project](https://github.com/traPtitech/traQ_S-UI/).
I digged down a bit and found that evaluating `workbox-build` was making it slow.

This PR changes `workbox-build` to be lazy loaded.
Since `workbox-build` is not always used, I think it is ok to do this.

Without this PR, my project takes about 1300 - 1500ms to start up.
With this PR, it takes about 500 - 600ms.

These values are obtained from output with `DEBUG="vite:config" npm run dev` like `bundled config file loaded in 565.03ms +0ms`.

### DIFF
--- a/src/modules.ts
+++ b/src/modules.ts
@@ -1,10 +1,19 @@
 import { resolve } from 'path'
 import { promises as fs } from 'fs'
-import { BuildResult, injectManifest, generateSW } from 'workbox-build'
+import type { BuildResult } from 'workbox-build'
 import { ResolvedConfig } from 'vite'
 import Rollup from 'rollup'
 import type { ResolvedVitePWAOptions } from './types'
 import { logWorkboxResult } from './log'
+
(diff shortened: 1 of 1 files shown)


userquin:
@sapphi-red 

sorry for the delay, testing with vitesse template also the dev startup from 2s to 1.25s, loading `workbox-build` on demand:

```shell
  vite v2.9.0 dev server running at:

  > Local: http://localhost:3333/
  > Network: use `--host` to expose

  ready in 1201ms.
```

with static import:
```shell
  vite v2.9.0 dev server running at:

  > Local: http://localhost:3333/
  > Network: use `--host` to expose

  ready in 2069ms.
```

userquin (approved):

PR #247, opened 2022-03-20
docs: fix default globPatterns

AllanChain:
Ref: https://developer.chrome.com/docs/workbox/reference/workbox-build/#property-GlobPartial-globPatterns

Using `**.{js,css,html}` will not cache any js or css file. It must be `**/*.{js,css,html}`


userquin:
@AllanChain weird, all workbox docs is using that string as the default value (at least on 6.4.2 version), I use the lib to extract the default value, it seems workbox update the docs:

imagen

userquin (approved):

AllanChain:
Yes, workbox's code was a little chaotic. But it's fixed now.

image

PR #261, opened 2022-04-08
fix: use all configured plugins but self with `injectManifest` (#260)

### LINKED ISSUE #260: `injectManifest` doesn't include enough plugins

```diff
  const includedPluginNames = [
+   "commonjs",
+   "vite:asset",
+   "vite:json",
    "alias",
    "vite:resolve",
    "vite:esbuild",
    "replace",
    "vite:define",
    "rollup-plugin-dynamic-import-variables",
    "vite:esbuild-transpile",
    "vite:terser"
  ];
```
  
keithlayne:
Fixes #260


userquin:
closing this, supersed by #262

PR #215, opened 2022-01-23
feat(inject manifest): add `commonjs` and `vite:json` plugins

schummar:
Hi, I use a custom service worker with the `'injectManifest'` strategy. When I try to import a commonjs module (e.g. fast-deep-equal), I receive the following error:
```
error during build:
Error: 'default' is not exported by node_modules/fast-deep-equal/index.js, imported by src/sw.ts
    at error (/home/projects/node-9idndn/node_modules/rollup/dist/shared/rollup.js:160:30)
    at Module.error (/home/projects/node-9idndn/node_modules/rollup/dist/shared/rollup.js:12429:16)
    at Module.traceVariable (/home/projects/node-9idndn/node_modules/rollup/dist/shared/rollup.js:12813:29)
    at ModuleScope.findVariable (/home/projects/node-9idndn/node_modules/rollup/dist/shared/rollup.js:11599:39)
    at Identifier.bind (/home/projects/node-9idndn/node_modules/rollup/dist/shared/rollup.js:6479:40)
    at CallExpression.bind (/home/projects/node-9idndn/node_modules/rollup/dist/shared/rollup.js:5087:23)
    at CallExpression.bind (/home/projects/node-9idndn/node_modules/rollup/dist/shared/rollup.js:9436:15)
    at CallExpression.bind (/home/projects/node-9idndn/node_modules/rollup/dist/shared/rollup.js:5083:31)
    at CallExpression.bind (/home/projects/node-9idndn/node_modules/rollup/dist/shared/rollup.js:9436:15)
    at ExpressionStatement.bind (/home/projects/node-9idndn/node_modules/rollup/dist/shared/rollup.js:5087:23)
```
StackBlitz that demonstrates the problem (just run `npm run build`): https://stackblitz.com/edit/node-9idndn?file=src/sw.ts

Vite itself can handle the module by default because it's using the commonjs rollup plugin. I found I can easily make it work in the service worker as well by adding `'commonjs'` to `includedPluginNames` and therefore allowing the PWA plugin to convert the module to esm.

Would you say it's reasonable to include it by default?
Any reason not to and if so how about the possibility to configure it?

PR #262, opened 2022-04-08
feat(injectManifest): allow configure Vite plugins

### LINKED ISSUE #260: `injectManifest` doesn't include enough plugins

userquin:
This PR includes:
- add `commonjs` and `vite:json`
- exports `defaultInjectManifestVitePlugins` from `constants.ts` module
- adds support to configure Vite plugins on `injectManifest` when building using Rollup

supersedes #215 
supersedes #261

closes #260

### DIFF
--- a/src/modules.ts
+++ b/src/modules.ts
@@ -2,8 +2,7 @@ import { resolve } from 'path'
 import { promises as fs } from 'fs'
 import { BuildResult, injectManifest, generateSW } from 'workbox-build'
 import { ResolvedConfig } from 'vite'
-// eslint-disable-next-line import/default
-import Rollup from 'rollup'
+import { rollup } from 'rollup'
 import type { ResolvedVitePWAOptions } from './types'
 import { logWorkboxResult } from './log'
 import { defaultInjectManifestVitePlugins } from './constants'
@@ -49,9 +48,7 @@ export async function generateInjectManifest(options: ResolvedVitePWAOptions, vi
     includedPluginNames.push(...defaultInjectManifestVitePlugins)
 
   const plugins = viteOptions.plugins.filter(p => includedPluginNames.includes(p.name))
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
-  const rollup = require('rollup') as typeof Rollup
-  const bundle = await rollup.rollup({
+  const bundle = await rollup({
     input: options.swSrc,
     plugins,
   })
--- a/src/modules.ts
+++ b/src/modules.ts
@@ -2,7 +2,6 @@ import { resolve } from 'path'
 import { promises as fs } from 'fs'
 import { BuildResult, injectManifest, generateSW } from 'workbox-build'
 import { ResolvedConfig } from 'vite'
-import { rollup } from 'rollup'
 import type { ResolvedVitePWAOptions } from './types'
 import { logWorkboxResult } from './log'
 import { defaultInjectManifestVitePlugins } from './constants'
@@ -48,6 +47,7 @@ export async function generateInjectManifest(options: ResolvedVitePWAOptions, vi
     includedPluginNames.push(...defaultInjectManifestVitePlugins)
 
   const plugins = viteOptions.plugins.filter(p => includedPluginNames.includes(p.name))
+  const { rollup } = await import('rollup')
   const bundle = await rollup({
     input: options.swSrc,
     plugins,
--- a/src/modules.ts
+++ b/src/modules.ts
@@ -5,6 +5,7 @@ import { ResolvedConfig } from 'vite'
 import Rollup from 'rollup'
 import type { ResolvedVitePWAOptions } from './types'
 import { logWorkboxResult } from './log'
+import { defaultInjectManifestVitePlugins } from './constants'
 
 export async function generateRegisterSW(options: ResolvedVitePWAOptions, mode: 'build' | 'dev', source = 'register') {
   const sw = options.base + options.filename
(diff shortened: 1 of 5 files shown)


src/modules.ts
```diff
@@ -36,19 +37,19 @@ export async function generateInjectManifest(options: ResolvedVitePWAOptions, vi
...
+  const includedPluginNames: string[] = []
+  if (typeof vitePlugins === 'function')
+    includedPluginNames.push(...vitePlugins(viteOptions.plugins.map(p => p.name)))
+  else
+    includedPluginNames.push(...vitePlugins)
+
+  if (includedPluginNames.length === 0)
+    includedPluginNames.push(...defaultInjectManifestVitePlugins)
+
+  const plugins = viteOptions.plugins.filter(p => includedPluginNames.includes(p.name)) as Plugin[]
```
keithlayne, line 50: BTW, why this type assertion? You're not actually importing the rollup Plugin type. You're asserting here that these are the deprecated browser plugin type from lib.dom.d.ts, which is weird and wrong. TS already knows that they are of the correct type.

src/modules.ts
```diff
@@ -2,9 +2,11 @@ import { resolve } from 'path'
 import { promises as fs } from 'fs'
 import { BuildResult, injectManifest, generateSW } from 'workbox-build'
 import { ResolvedConfig } from 'vite'
+// eslint-disable-next-line import/default
 import Rollup from 'rollup'
```
keithlayne, line 6:
```suggestion
import { rollup } from 'rollup'
```

src/modules.ts
```diff
@@ -36,19 +38,19 @@ export async function generateInjectManifest(options: ResolvedVitePWAOptions, vi
...
+    includedPluginNames.push(...vitePlugins(viteOptions.plugins.map(p => p.name)))
+  else
+    includedPluginNames.push(...vitePlugins)
+
+  if (includedPluginNames.length === 0)
+    includedPluginNames.push(...defaultInjectManifestVitePlugins)
+
+  const plugins = viteOptions.plugins.filter(p => includedPluginNames.includes(p.name))
   // eslint-disable-next-line @typescript-eslint/no-var-requires
   const rollup = require('rollup') as typeof Rollup
```
keithlayne, line 53:
```suggestion
```

src/modules.ts
```diff
@@ -36,19 +38,19 @@ export async function generateInjectManifest(options: ResolvedVitePWAOptions, vi
...
+  if (typeof vitePlugins === 'function')
+    includedPluginNames.push(...vitePlugins(viteOptions.plugins.map(p => p.name)))
+  else
+    includedPluginNames.push(...vitePlugins)
+
+  if (includedPluginNames.length === 0)
+    includedPluginNames.push(...defaultInjectManifestVitePlugins)
+
+  const plugins = viteOptions.plugins.filter(p => includedPluginNames.includes(p.name))
   // eslint-disable-next-line @typescript-eslint/no-var-requires
```
keithlayne, line 52:
```suggestion
```

src/modules.ts
```diff
@@ -36,19 +38,19 @@ export async function generateInjectManifest(options: ResolvedVitePWAOptions, vi
...
-    'vite:resolve',
-    'vite:esbuild',
-    'replace',
-    'vite:define',
-    'rollup-plugin-dynamic-import-variables',
-    'vite:esbuild-transpile',
-    'vite:terser',
-  ]
-  const plugins = viteOptions.plugins.filter(p => includedPluginNames.includes(p.name)) as Plugin[]
   const bundle = await rollup.rollup({
```
keithlayne, line 54:
```suggestion
  const bundle = await rollup({
```

keithlayne (commented):
Sorry, had to do line by line, but these changes should fix the import weirdness here.

userquin:
@keithlayne the original was from Vite:
- https://github.com/vitejs/vite/blob/main/packages/vite/src/node/plugins/worker.ts#L5
- https://github.com/vitejs/vite/blob/main/packages/vite/src/node/plugins/worker.ts#L76

keithlayne:
```ts
import type Rollup from 'rollup'
...
  const rollup = require('rollup') as typeof Rollup
```

One difference is the `import type`. But the next line doesn't even make any sense - `typeof` is only applied to values, and that's a type. I'd suspect that TS would create a compiler error there.

But...yeah. They're doing it wrong :stuck_out_tongue:

userquin:
@keithlayne I'll revert last commit adding `type` on the import, we need to load rollup on demand, we don't need it when using `generateSW` strategy

keithlayne:
In that case use `const { rollup } = await import('rollup)`.

userquin:
@keithlayne just testing some examples on this repo and on vitesse clone on esm and cjs environments

keithlayne:
That looks way better to me!  In general I just expect whatever your transpiler is (esbuild?) to just Do The Right Thing :tm:

userquin:
Maybe we need to move to unbuild... the generated mjs uses require and throws if missing

keithlayne:
> Maybe we need to move to unbuild... the generated mjs uses require and throws if missing

wut. 

I'm not familiar with `unbuild`. Seems overly fancy. What are the goals for publishing? I'd think just using `tsc` would probably work great. Using something like `api-extractor` would be easy too. Notably, I didn't see any typechecking in your build process, which is something that's important to me personally.

keithlayne:
Honestly it makes no sense to me that any tool in ESM mode would actually transform `import()` to `require`. That's gotta be a bug. (Or a config issue, but your config looks as simple as can be)

keithlayne:
I saw @antfu 's blog post (https://antfu.me/posts/publish-esm-and-cjs) about `tsup` and `unbuild`, so if `unbuild` is an easy switch, looks like something they prefer and appears to be able to do the job. Probably needs that `cjsBridge` setting, I saw at least one use of `__dirname`.

keithlayne:
Actually, I just looked at the original JS in my local patch, and the ESM build was already using that `__require` helper - are you sure that it's really broken? I can't easily confirm, when I run vite it's loading the CJS version.

userquin:
@keithlayne it is working using require or dynamic import.

I have it also working on local using unbuild with dynamic import but only the vite plugin, clients built with tsup since we only need them to load using read file from the plugin.

userquin:
@keithlayne we are using `tsup`, its target is `node12` (the default,  just check the logs when building) and this is why it is generating the `__toModule(__require(name))`, we should use something like the following (or use `local-pkg`, see https://github.com/antfu/local-pkg#usage, last line using `importModule`):
```ts
import { createRequire } from 'module'

const _require = typeof __dirname === 'undefined' ? createRequire(import.meta.url) : require
```

For context: https://github.com/antfu/unplugin-icons/pull/71#issuecomment-927814752

userquin:
here the patch with the changes: 
[unbuild.zip](https://github.com/antfu/vite-plugin-pwa/files/8456717/unbuild.zip)

userquin:
here the full version also including #220 :
[unbuild-full-2.zip](https://github.com/antfu/vite-plugin-pwa/files/8456775/unbuild-full-2.zip)

antfu (approved):

PR #300, opened 2022-06-06
feat: use `webManifestUrl` on `injectDevManifest`

userquin:
I forgot to use `webManifestUrl` also on `injectDevManifest`: if someone using it will not work (`SvelteKit`  will not fail since the `manifest.webmanifest` is included on the `_layout`)

### DIFF
--- a/src/dev.ts
+++ b/src/dev.ts
@@ -58,6 +58,8 @@ export async function loadDev(id: string, options: ResolvedVitePWAOptions, viteC
         // the sw precache (self.__SW_MANIFEST) will be empty since we're using `dev-dist` folder
         // we only need to add the navigateFallback if configured
         const navigateFallback = options.workbox.navigateFallback
+        // we need to exclude the manifest.webmanifest from the sw precache: avoid writing it to the dev-dist folder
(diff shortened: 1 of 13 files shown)


userquin:
I need to add the content type when serving it, I'll fix it.

src/dev.ts
```diff
@@ -58,13 +58,16 @@ export async function loadDev(id: string, options: ResolvedVitePWAOptions, viteC
...
+        // we need to exclude the manifest.webmanifest from the sw precache: avoid writing it to the dev-dist folder
+        const webManifestUrl = options.devOptions.webManifestUrl ?? `${options.base}${options.manifestFilename}`
         const { filePaths } = await generateServiceWorker(
           Object.assign(
             {},
             options,
             {
               workbox: {
                 ...options.workbox,
+                navigateFallbackDenylist: [new RegExp(`^${webManifestUrl}$`)],
```
antfu, line 70: Do we need to escape regex here?

userquin: no, using the constructor will replace with the escaped literal: `/manifest.webmanifest` => `^\/manifest.webmanifest$`.

I check it.

userquin: wait, I can double test...

userquin: It is ok:

imagen

PR #301, opened 2022-06-06
fix: use `esno 0.14.1`

userquin:
Until https://github.com/vuejs/vitepress/pull/709 merged/released on `Vitepress` we'll need old version of `esno` to use it on docs build


userquin:
ok, I need to change it on local to rebuild the `lock` file.

userquin:
@antfu moved to #300

PR #306, opened 2022-06-11
fix: skip waiting check

mastercoms:


userquin:
@mastercoms thanks a lot, looks like you've found the ghost of `injectRegister`

userquin:
@mastercoms can you change also this code (replace the `transformIndexHtml`) to `src/index.ts` module?
```ts
      transformIndexHtml: {
        enforce: 'post',
        transform(html) {
          if (options.disable)
            return html

          // if virtual register is requested, do not inject.
          if (options.injectRegister === 'auto')
            options.injectRegister = useImportRegister ? null : 'script'

          return injectServiceWorker(html, options)
        },
      },
```

userquin (approved):

userquin:
to test it use the cdn example, just remove the `base` and add the `registerType: 'autoUpdate'` on the vite config file and remove the `ReloadPrompt` from the `App`

PR #311, opened 2022-06-23
fix: support empty base in vite options

nornagon:
vite supports an [empty string](https://vitejs.dev/config/#base) for `base`, but if `base: ''` then VitePWA injects a manifest link to `/manifest.webmanifest`, instead of `manifest.webmanifest`. This fixes that :)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -5,7 +5,7 @@ export function slash(str: string) {
 export function resolveBathPath(base: string) {
   if (isAbsolute(base))
     return base
-  return !base.startsWith('/') && !base.startsWith('./')
+  return base && !base.startsWith('/') && !base.startsWith('./')
     ? `/${base}`
     : base
 }


userquin:
@nornagon in `Vite 2`, using base with empty string, the result can be weird/indeterminate, you should use `./` instead; in `Vite 3` you can use `base : ''` or `base: './'` to configure relative base

We'll keep in hold state...

thx

antfu (approved):

userquin:
We will wait for the release of Vite 3, we need to review everything related to `base`, we are using it to set many defaults, such as `scope` and we are not sure yet if only including the changes in this PR will make the plugin work correctly.

userquin:
To support **Vite 3 Relative Base**, we need to modify the code here to resolve the `scope` when registering the service worker, via the inline/registerSW/virtual scripts: for example, we can use [document.baseURI](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base), but this requires the user to modify the entry points after the app is deployed by adding `<base>`.

The webmanifest also has the above problem in that its 'scope' cannot be changed once generated: of course the user must also modify the webmanifest once it is deployed.

I think support for this can't be added here, it's a limitation on how the service worker works: if a user is going to deploy their application to a bunch of different `base` servers, they must be known and the user should create a `build` for each of them with the corresponding configuration, changing the `base`.

PR #315, opened 2022-06-30
docs: add twitter stuff

userquin:


userquin:
@antfu The preview not working on twitter, maybe we need to add the `twitter:url` entry or maybe due to the crawler: pasting the url on a tweet should show the `twitter:image`, but it doesn't, I'll check it later when docs published with latest changes on the home.

userquin:
here https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary the twitter entries are using meta with name, on vitejs docs are using meta with property  (https://github.com/vitejs/vite/blob/main/docs/.vitepress/config.ts#L18)

docs/index.md
```diff
@@ -5,11 +5,8 @@ sidebar: false
 title: Vite Plugin PWA
 
 hero:
-  name: PWA
-  text: Vite Plugin
-  tagline: |
-    Zero-config and framework-agnostic
-    PWA Plugin for Vite
+  name: Vite Plugin PWA
+  text:  Zero-config and framework-agnostic PWA Plugin for Vite
```
antfu, line 9: I still prefer the old texts

userquin: ok

PR #316, opened 2022-06-30
Add support for display_override

michaelhue:
This change adds type support for the [`display_override`](https://developer.mozilla.org/en-US/docs/Web/Manifest/display_override) manifest member.


userquin (approved):

userquin:
@michaelhue thx, just in time

PR #321, opened 2022-07-02
fix: register service worker in dev

userquin:
This PR also includes:
- fix linting,  excluded a few folders for dev and build
- included `prev/next` links ready for next `VitePress alpha` release: https://vitepress.vuejs.org/guide/theme-prev-next-link.html 
- update deps with `taze -r -w`: beware, `taze` is resolving `VitePress` versions wrong, resolving to `draft` instead to `alpha`
- split plugins

EDIT: there is no race condition, `useImportRegister` never updated on dev with dev options enabled:
https://github.com/antfu/vite-plugin-pwa/blob/main/src/index.ts#L133

supersedes #319 (sw in dev part)


userquin:
@antfu I'll merge into main, you can release it or wait until I finish the new PR about adding support also for unregister the sw/pwa via `selfDestroying` option

PR #332, opened 2022-07-11
docs: include `allowlist` hints in development entry

userquin:


userquin:
@antfu I'll add some new options to the devOptions about this, just to avoid the pitfall, DON'T MERGE (on draft: I'll remove this and update from a new one PR including the logic and the docs)

userquin:
supersed by #335 (docs part)

PR #340, opened 2022-07-17
Fix dev-dist location

SteveALee:
### Description

The `dev-dist` folder is created relative to the vite `root` setting. Often this is fine but if you want a more typical directory structure by setting `root` to be `src` and `outDir` to be `../dist` then `dev-dist` appears under `src`. not good!

This PR changes the location of `dev-dist` to be a sibling of `outDir`

I tested by hacking `node-modules/vite-plugin-pwa` in https://github.com/music-practice-tools/musical-parameters with my config and the default config.

I think this is a safe change. The only break might be dev-dist moving and so not matching .gitignore

### Linked Issues

None

### Additional context

None

### DIFF
--- a/src/plugins/dev.ts
+++ b/src/plugins/dev.ts
@@ -105,7 +105,7 @@ export function DevPlugin(ctx: PWAPluginContext): Plugin {
           return undefined
         }
         if (id.endsWith(swDevOptions.swUrl)) {
-          const globDirectory = resolve(viteConfig.root, viteConfig.outDir, '..',  'dev-dist')
+          const globDirectory = resolve(viteConfig.root, viteConfig.build.outDir, '..',  'dev-dist')
           if (!existsSync(globDirectory))
             mkdirSync(globDirectory)
 
@@ -158,7 +158,7 @@ export function DevPlugin(ctx: PWAPluginContext): Plugin {
 
 async function createDevRegisterSW(options: ResolvedVitePWAOptions, viteConfig: ResolvedConfig) {
   if (options.injectRegister === 'script') {
-    const devDist = resolve(viteConfig.root, viteConfig.outDir, '..', 'dev-dist')
+    const devDist = resolve(viteConfig.root, viteConfig.build.outDir, '..', 'dev-dist')
     if (!existsSync(devDist))
       mkdirSync(devDist)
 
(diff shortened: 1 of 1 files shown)


userquin:
@SteveALee why using the outDir if we have the root folder?

SteveALee:
I may be missing understanding votes intention but I think root is for the src code. Plonking a generated folder under that is just weird and confusing, esp as is git ignored.

Maybe the vite team don't think like me but I always put all src in /src leaving project root for config. Having config and src all mixed just looks a mess to me.

Many other projects do similar.

userquin:
@SteveALee you're mixing 2 distinct concepts, the root is the root folder (where the package.json is located), you can use `/` to reference the root folder ando so in the entry point you need to use `/src/main.ts`.

In your example (email), if the root folder was `src`, then the `/src/main.ts` in the entry point should be `/main.ts`: Vite is resolving it correctly using `/src/main.ts`.

SteveALee:
Hmm, that's not what I find. Project root is as usual and and vite root is where the src code lives, not package json

I set vite root to /src and dev-dist gets created under it, Which is what I do not want

```javascript
export default defineConfig({
  root: 'src',
  publicDir: "../public",
  build: {
    outDir: '../dist',  // note that dev-dist always appears under src/ :()
    emptyOutDir: true
  },
```

and with my fix

image

Anyway, we may be talking at crossed purposes but I can;t spend more time. So feel free to ignore.

Thanks

userquin:
@SteveALee if you change the root folder you should also change the rest of values: maybe you can change this PR to allow configure the `dev-dist` folder, just add it in the `devOptions`, but I cannot tell you if it will work.

You're changing the layout of Vite, and so you have to deal with that change: a lot of plugin can stop working.

SteveALee:
@userquin 

>In your example (email), if the root folder was src, then the /src/main.ts in the entry point should be /main.ts: Vite is resolving it correctly using /src/main.ts.

I finally got what you are saying. But why would /src/main.ts not map to /src/main.ts ? What about /src/subdir/main.ts? How can you manage the structure of the target PWA if it flattens everything?

I took vite root to be the src root not project root as that is in my experience always the repo root. Maybe you can set it for something like monorepos which I've never used?

I'm obviously missing something here. 

I also realised my testing was void due to PWA caching :(

I guess There are other source layouts in vite  land to those I've always used- ie proj root is for proj config only, everything in own folders. For example I'd like all those fav and manifest icons to be in own folder.

Thanks for your time. I'll close and live with the apparent miss match for now.

SteveALee:
@userquin  I tried using the typical config but it seem vite always looks for index.html in the root :( At least I could not find a way to change that.

PR #341, opened 2022-07-18
docs(fix): lang and copy in code blocks in light mode

userquin:
Just compare current code blocks with the preview.


antfu:
Maybe better to be improve in VitePress directly?

PR #345, opened 2022-07-25
fix: respect base path in virtual development SW

wvffle:
### Description

In some cases, projects are behind a reverse proxy even in dev mode. The virtual dev module should respect the `base` option from vite config.


userquin (approved):

wvffle:
Is there anything blocking this PR?

PR #356, opened 2022-08-17
feat: add VitePress integration

### LINKED ISSUE #355: add VitePress integration

Finally VitePress has been released with the `buildEnd` hook: we can now create an integration here using the code in the VitePress PR: https://github.com/vuejs/vitepress/pull/709#issuecomment-1146139835

### Suggested solution

Just add a new integration for VitePress: add a factory function to create the `async buildEnd` hook.

I need to check the logic for this integration!.

### Alternative

_No response_

### Additional context

_No response_

### Validations

userquin:
I need to change the docs

closes #355


src/integrations/vitepress/index.ts
```diff
@@ -0,0 +1,20 @@
+import { resolveConfig } from 'vite'
+import type { VitePWAOptions, VitePluginPWAAPI } from '../../types'
+import { VitePWA } from '../../index'
+
+export function createBuildEndHook(userOptions: Partial<VitePWAOptions>): (siteConfig: any) => Promise<void> {
+  return async (siteConfig: any) => {
+    // run first integration hook
+    await userOptions.integrationHook?.(siteConfig)
```
userquin, line 8: I'll remove that, ppl can do it.

userquin: @antfu Finally I have left this option included

src/integrations/vitepress/index.ts
```diff
@@ -0,0 +1,20 @@
...
+    const viteConfig = await resolveConfig({
+      plugins: [VitePWA(userOptions)],
+    },
+    'build',
+    'production',
```
userquin, line 13: should these 2 options be configurable?

antfu:
Can't we await Vite 3.1 for the object hook? I am still not a fan of having integration specific logic built in.

userquin:
> Can't we await Vite 3.1 for the object hook? I am still not a fan of having integration-specific logic built in.

We need to do this way for SvelteKit and VitePress, both repos reject the integration.

For VitePress there is no race condition we only need to use the build end hook 

For SvelteKit we need both, the integration and the sequential hook.

An alternative can be split the repo similar to UnoCSS with separate integrations.

docs/.vitepress/config.ts
```diff
@@ -227,6 +308,7 @@ function prepareSidebar(idx: number) {
 const ogUrl = 'https://vite-plugin-pwa.netlify.app/'
 const ogImage = 'https://vite-plugin-pwa.netlify.app/og-image.png'
 
+// @ts-expect-error avoid TS2321: Excessive stack depth comparing types
```
userquin, line 311: I need to remove this line, it seems on IntelliJ is not working but it works in VSCode.

userquin: ok, this error seems to appear and dissapear randomly, so we can keep it: https://discord.com/channels/325477692906536972/1012326745753530419/1012347386653319248

userquin:
included in the integration repo

PR #361, opened 2022-08-21
fix(dev): type module and custom sw not working when not using virtual modules

### LINKED ISSUE #359: `registerSW.js` 404s in dev mode with `injectManifest` strategy
### What am I trying to do

I am trying to get the service worker to register in dev mode with `strategies = injectManifest` and `devOptions.enabled = true`

### What do I expect to happen

I expect the service worker to load correctly in dev mode.

### What happens instead

The service worker registration does not take place as `registerSW.js` cannot be found:

image

Running `pnpm build` and then serving the production bundle works without issues, and the service worker successfully registers.

### Reproduction

https://github.com/lhjt/issue-repro

### System Info

```Shell
System:
    OS: macOS 13.0
    CPU: (10) arm64 Apple M1 Max
    Memory: 5.69 GB / 64.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
    npm: 8.11.0 - ~/.nvm/versions/node/v16.16.0/bin/npm
  Browsers:
    Chrome Canary: 107.0.5252.0
    Firefox Developer Edition: 103.0
    Safari: 16.0
```

### Used Package Manager

pnpm

### Validations

- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided reproduction is a [minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

ARYX01:
### Description

devOptions.type: 'module' now using it for generating inside the registerSW.js

### Linked Issues
closes #359

### Additional context

The dev-dist folder is unreachable when running dev vite3.

I suggest not creating this folder and instead add the files inside the ws /@vite-plugin-pwa/ like the pwa-entry-point-loaded and point there the registerPath


src/html.ts
```diff
@@ -3,10 +3,10 @@ import type { ResolvedVitePWAOptions } from './types'
 
 export function generateSimpleSWRegister(options: ResolvedVitePWAOptions, dev: boolean) {
   const path = dev ? `${options.base}${DEV_SW_NAME}` : `${options.base}${options.filename}`
-
+  const swType = options.devOptions.type ? options.devOptions.type : "classic";
```
userquin, line 6: move this const inside the if statement

src/plugins/dev.ts
```diff
@@ -79,8 +81,10 @@ export function DevPlugin(ctx: PWAPluginContext): Plugin {
       return undefined
     },
     async load(id) {
-      if (id === DEV_SW_VIRTUAL)
+      if (id === DEV_SW_VIRTUAL || id === `${ctx.options.base}${DEV_SW_VIRTUAL}`){
+	    console.log('load',id)
```
userquin, line 85: remove this line

src/plugins/dev.ts
```diff
@@ -60,8 +60,10 @@ export function DevPlugin(ctx: PWAPluginContext): Plugin {
       }
     },
     resolveId(id) {
-      if (id === DEV_SW_VIRTUAL)
+      if (id === DEV_SW_VIRTUAL || id === `${ctx.options.base}${DEV_SW_VIRTUAL}`){
+	    console.log('resolveId',id)
```
userquin, line 64: remove this line

userquin:
@ARYX01 we use `pnpm`as package manager: please remove the package-lock-json file

userquin:
@antfu it is weird, Vite 2 seems to work properly since it is forcing to use a base (by default `/`): internally the prefix is being removed and so this lines will work:
- the virtual path fot the websobkect ping/pong: https://github.com/antfu/vite-plugin-pwa/blob/main/src/plugins/dev.ts#L15
- this script will have double starting slash: https://github.com/antfu/vite-plugin-pwa/blob/main/src/plugins/dev.ts#L40
- Vite 2 is working properly here: https://github.com/antfu/vite-plugin-pwa/blob/main/src/plugins/dev.ts#L63
- Vite 2 is working properly also here: https://github.com/antfu/vite-plugin-pwa/blob/main/src/plugins/dev.ts#L82

The same code with Vite 3 will not work, it seems it is not removing one slash: I think it is because th target project is not using the `base` entry, and so the plugin will force to use the base `/ `, then https://github.com/antfu/vite-plugin-pwa/blob/main/src/plugins/dev.ts#L40 will have 2 slashes, and so the plugin will not resolve that url.

This PR will just remove the `/`  from the constant, I need to check it with Vite 2: https://github.com/antfu/vite-plugin-pwa/pull/361/files#diff-8fa4b52909f895e8cda060d2035234e0a42ca2c7d3f8f8de1b35a056537bf199R11.

userquin:
I will change the `transformIndexHtml` hook on dev.ts to include directly the HMR ping/pong content to the script  adding a small timeout before the ping, and so we dont need to deal with the base.

src/plugins/dev.ts
```diff
@@ -37,7 +38,12 @@ export function DevPlugin(ctx: PWAPluginContext): Plugin {
 
         return html.replace(
           '</body>',
-            `<script type="module" src="${options.base}${DEV_SW_VIRTUAL}"></script></body>`,
+            `
+<script type="module">
```
userquin, line 42: should we check CORS here?

antfu: We could leave it for now, and bring it back if there is really a valid case.

ARYX01:
After adding an reload prompt component that uses `import { useRegisterSW } from 'virtual:pwa-register/svelte'` and setting pwaOptions `registerType: 'prompt'`, the /dev-dist/registerSW.js is not created anymore.
So before was working only with `injectRegister: 'inluine'`, but now it works also with `injectRegister: 'script'`.

userquin:
we only need to fix the SSR: #364

antfu (approved):

PR #365, opened 2022-08-31
fix: escape swType when generating registration file

rdunk:
### Description

The latest version `v0.12.4` added some dev specific logic in the `generateSimpleSWRegister` method to generate the `registerSW.js` file.

The `swType` isn't wrapped in quotation marks, so you see an error, e.g. `Uncaught ReferenceError: classic is not defined`.

This PR just adds the quotation marks.

### DIFF
--- a/src/html.ts
+++ b/src/html.ts
@@ -7,7 +7,7 @@ export function generateSimpleSWRegister(options: ResolvedVitePWAOptions, dev: b
   // we are using HMR to load this script: DO NOT ADD window::load event listener
   if (dev) {
     const swType = options.devOptions.type ?? 'classic'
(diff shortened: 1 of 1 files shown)


userquin:
@rdunk in the meantime just use `auto` instead `inline`.

rdunk:
@userquin Thanks for the suggestion, but I'm not using `inline` as far as I'm aware? Just tried explicitly setting `injectRegister: 'auto'` but still see the same error appearing in dev mode.

userquin:
@rdunk uhmmm, too many things man ;) , you're right, just try inline, this will bypass the call to create the `registerSW.js`

PR #367, opened 2022-09-02
feat: load workbox window with dynamic import

### LINKED ISSUE #354: Load workbox-window with dynamic import

Instead using static import on register service worker client module use dynamic import.

Check if we can use the mjs module from workbox-window (source).

### Suggested solution

This will make the client load fast.

### Alternative

_No response_

### Additional context

https://developer.chrome.com/docs/workbox/modules/workbox-window/#loading-workbox-with-javascript-bundlers

### Validations

userquin:
closes #354

### DIFF
--- a/src/client/build/register.ts
+++ b/src/client/build/register.ts
@@ -25,9 +25,11 @@ export function registerSW(options: RegisterSWOptions = {}) {
 
   let wb: import('workbox-window').Workbox | undefined
   let registration: ServiceWorkerRegistration | undefined
+  let registerPromise: Promise<void>
   let sendSkipWaitingMessage: () => Promise<void> | undefined
 
   const updateServiceWorker = async (reloadPage = true) => {
+    await registerPromise
     if (!auto) {
       // Assuming the user accepted the update, set up a listener
       // that will reload the page as soon as the previously waiting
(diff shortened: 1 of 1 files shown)


antfu (approved):

src/client/build/register.ts
```diff
@@ -39,58 +41,66 @@ export function registerSW(options: RegisterSWOptions = {}) {
...
-      }
+      await sendSkipWaitingMessage?.()
     }
   }
 
-  if ('serviceWorker' in navigator) {
-    // __SW__, __SCOPE__ and __TYPE__ will be replaced by virtual module
-    wb = new Workbox('__SW__', { scope: '__SCOPE__', type: '__TYPE__' })
+  async function register() {
+    const { Workbox, messageSW } = await import('workbox-window')
```
userquin, line 49: Move the import inside the if statement

strarsis:
@userquin: This may be a bit on a tangent - but how can I use `workbox-window`, notably the `messageSW` method from inside the page/window? I import and invoke `registerSW` from `virtual:pwa-register`. But `workbox` is initialized in the generated code and not attached anyway (e.g. `window`), so I can use it.

userquin:
@strarsis check this https://developer.chrome.com/docs/workbox/modules/workbox-window/#advanced-bundling-concepts

You can use static/dynamic import, isn't working for you?: `import { messageSW } from 'workbox-window'`

PR #376, opened 2022-09-17
feat: virtual module for PWA info

### LINKED ISSUE #375: expose PWA info

Sometimes we need some info:
- dev environment: to allow configure sw properly (will be also ported to Vite 3.1 PR: SvelteKit will require it)
- webmanifest info: expose url and `use-credentials`
- service worker info: expose sw registration info
- service worker url on registered callback: to allow useit with periodic sw updates (with dev options enable the name is not the filename, it is the fixed name defined in constants.ts module)

### Suggested solution

Expose it using the Rollup Plugin Intercommunication API.

### Alternative

None

### Additional context

Just read the description.

### Validations

userquin:
This PR also includes:
- all examples updated to use the new `onRegisteredSW` callback option
- deprecates `onRegistered` callback
- update docs (wip) 

This PR should go to current main branch.

closes #375

### DIFF
--- a/src/plugins/info.ts
+++ b/src/plugins/info.ts
@@ -1,0 +1,60 @@
+import type { Plugin } from 'vite'
+import type { PWAPluginContext } from '../context'
+import {
+  PWA_INFO_VIRTUAL,
+  RESOLVED_PWA_INFO_VIRTUAL,
+} from '../constants'
+import type { VitePluginPWAAPI } from '../types'
+
+export function InfoPlugin(ctx: PWAPluginContext, api: VitePluginPWAAPI) {
+  return <Plugin>{
+    name: 'vite-plugin-pwa:info',
+    enforce: 'post',
+    resolveId(id) {
+      if (id === PWA_INFO_VIRTUAL)
+        return RESOLVED_PWA_INFO_VIRTUAL
+
+      return undefined
+    },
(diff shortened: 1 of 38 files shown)


userquin:
@antfu I'm porting this to the Vite 3.1 PR and will check if there is something missing we still need to add.

userquin:
I need to review also last commit on my local...

userquin:
I'll try to port this tmr to Vite 3.1 PR and test it with kit integration

src/plugins/info.ts
```diff
@@ -0,0 +1,60 @@
...
+
+function generatePwaInfo(ctx: PWAPluginContext, api: VitePluginPWAAPI) {
+  const webManifestData = api.webManifestData()
+  if (!webManifestData)
+    return 'export const pwaInfo = undefined;'
+
+  const { href, useCredentials, toLinkTag } = webManifestData
+  const registerSWData = api.registerSWData()
+
+  const entry = {
```
antfu, line 34: Can we expose a type for it?

antfu (approved):

src/plugins/info.ts
userquin: I'll add the type in main branch, I have errors when filling it, the type will be internal, just a copy paste from the virtual

PR #383, opened 2022-09-27
docs: update Astro integration

userquin:
don't merge until we publish Astro integration


userquin:
included in the new docs

PR #390, opened 2022-10-09
feat: custom sw without injection point

### LINKED ISSUE #385: allow use custom sw without injection point
This will allow ppl to NOT include `self.__WB_MANIFEST` in their custom sw.

The fix is quite simple, just return without calling workbox before this line if there is no `injectionPoint`: 
https://github.com/vite-pwa/vite-plugin-pwa/blob/main/src/modules.ts#L121

userquin:
closes #385


userquin:
@antfu Added simple example with a default preset for navigation, assets and images: it is the default from Nuxt (I'm working also to include an integration module for Nuxt).

I need to add some docs, uing the sw this way the custom sw should be versioned to detect changes: in the example I just include a timestamp. I think using it with prompt makes no sense, that's why Nuxt always force client claims on the sw template (auto update strategy here).

PR #394, opened 2022-10-14
fix(client): multiple clients with prompt for update

### LINKED ISSUE #393: `updateServiceWorker(true)` doesn't reload other open tabs
Hello! Thank you for this plugin. I've noticed there is an issue with the reload prompt and how the `controlling` event is being listened for. According to the Workbox example [here](https://developer.chrome.com/docs/workbox/handling-service-worker-updates/), the reload listener `wb.addEventListener('controlling', () => { window.location.reload(); });` is added _before_ the prompt is shown.

In the vite-plugin-pwa React [example](https://vite-plugin-pwa.netlify.app/frameworks/react.html#prompt-for-update), the listener is only added when `updateServiceWorker(true)` is called on the action where a user clicks reload on the reload prompt.

This means if I have two tabs open, and a service worker update is published, both tabs show the reload prompt, however clicking reload on the first tab only reloads the first tab, then clicking reload on the second tab does nothing (as the new SW has already activated).

Clicking reload on the second tab _does_ add the event listener, as the event listener is added when `updateServiceWorker(true)` is called by clicking the reload button in the prompt. This means, if another service worker update is published, both tabs show a reload prompt, and clicking reload on the first tab, also reloads the second tab.

### Reproduction steps:

1. Build react example app with service worker
2. Serve the app and ensure SW is active
3. Open additional tab with the same app
4. Make a code change and rebuild
5. Observe reload prompt in both tabs
6. Click reload prompt in first tab
7. Observe first tab **does** reload, second tab **does** not reload
9. Click reload prompt in second tab
10. Observe first tab **does not** reload, second tab **does not** reload
11. Make a code change and rebuild
12. Observe reload prompt in both tabs
13. Click reload prompt in first tab
14. Observe first tab **does** reload, observe second tab **does** reload

### Suggested fix

The event listener could be added during the `onNeedRefresh` call, not during the `updateServiceWorker` action, to ensure other tabs also have the event listener added.

### Current workaround

Currently I've made the following change in my project. The `useEffect` ensures the event listener is added once the `ReloadPrompt` is rendered for each tab, then calling `updateServiceWorker(false)` just sends the `SKIP_WAITING` message.

https://github.com/vite-pwa/vite-plugin-pwa/compare/main...develohpanda:vite-plugin-pwa:patch-1

I would be happy to raise a PR if you could guide me on the approach you'd like to use!

userquin:
I have tested this PR using `pnpm run build && pnpm run examples` using vue example with 2 chrome instances, first one with 2 tabs. Once I refresh one app, the reload prompt will be shown on other tab/instance and once reload the app, the 3 clients are updated (the prompt dialog also gone on the 3 clients).

EDIT: once installed on the 3 clients, stop the server and run `pnpm run examples` again with the same configuration and refresh the page on 1 client.

We need to test it on Firefox and Safari on others OS: only tested on Windows 10 PRO.

/cc @develohpanda 

closes #393

### DIFF
--- a/client.d.ts
+++ b/client.d.ts
@@ -20,6 +20,12 @@ declare module 'virtual:pwa-register' {
     onRegisterError?: (error: any) => void
   }
 
+  /**
+   * Registers the service worker returning a callback to reload the current page when an update is found.
+   *
+   * @param options the options to register the service worker.
+   * @return (reloadPage?: boolean) => Promise<void> From version 0.13.2+ `reloadPage` param is not used anymore.
+   */
   export function registerSW(options?: RegisterSWOptions): (reloadPage?: boolean) => Promise<void>
 }
 
@@ -52,6 +58,11 @@ declare module 'virtual:pwa-register/vue' {
   export function useRegisterSW(options?: RegisterSWOptions): {
     needRefresh: Ref<boolean>
     offlineReady: Ref<boolean>
+    /**
+     * Reloads the current window to allow the service worker take the control.
+     *
+     * @param reloadPage From version 0.13.2+ this param is not used anymore.
+     */
     updateServiceWorker: (reloadPage?: boolean) => Promise<void>
   }
 }
@@ -85,6 +96,11 @@ declare module 'virtual:pwa-register/svelte' {
   export function useRegisterSW(options?: RegisterSWOptions): {
     needRefresh: Writable<boolean>
     offlineReady: Writable<boolean>
+    /**
+     * Reloads the current window to allow the service worker take the control.
+     *
+     * @param reloadPage From version 0.13.2+ this param is not used anymore.
+     */
     updateServiceWorker: (reloadPage?: boolean) => Promise<void>
   }
 }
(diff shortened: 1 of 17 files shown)


userquin:
it seems to be related with the heuristic algorithm used in workbox-window: I'll try to test it awaiting 1 minute between updates

userquin:
After digging in the workbox window source code I've fixed the problem: we can also remove all warnings about heuristic algorithm from the docs

`workbox-window` tries to solve the problem when the app is opened in multiple clients changing some internal listeners for `updatefound` event from the sw. 
When the sw that fires the update is not the current one, some params on the event fired on `installed` event will change, for example, `isUpdate` will be false and `isExternal` will be true (on the client that fires the update `isUpdate` will be true and `isExternal` will be false.
We only need to change the `installed` listener to check if `!evt.isUpdate && evt.isExternal`, ths way any prompt will not be shown and we just reload the window.
If it is `ect.isUpdate` then we just reload the page only when using `autoUpdate` and finally if the sw is not `selftDestroying` we show the ready to work offline prompt.

Issue #396, opened 2022-10-15
Vite PWA not including correct path to images in manifest


RDjarbeng:
I have a site with PWA img assets in 

img/icons/imagename.png

I am trying to build the site with vite and vite-plugin-pwa to use it as a pwa. 
The vite.config.js and the index.html are in the project root directory. Images are in img directory.

The problem is that when I run the build file places the imgs in the folder `assets/imageName###.png`
For example `chrome192.png` is placed in `assets/chrome192.f25426fd.png`
However, the manifest file generated upon build `manifest.webmanifest` still contains 

            src: 'img/icons/chrome192.png',

The application tab for dev tools in chrome shows `{rootURL}/img/icon/chrome192.png` not found. Which is expected since the bundling with vite build places it in a different folder (assets). 

Why does it not update the path of the images in the generated manifest.webmanifest? 
Isn't that the whole point of the vite-plugin-pwa to keep track of the filenames that change upon build. 

Another issue is that I have different routes eg: html/about 
Inside the about.html generated on build, the web manifest path is given as:

    <link rel="manifest" href="./manifest.webmanifest">

It uses this path instead of using `../manifest.webmanifest` or maybe a path from the root without the ./ such as `href="./manifest.webmanifest"`

My vite.config.js is shown below

    import { resolve } from 'path';
    import { defineConfig } from 'vite';
    import { VitePWA } from 'vite-plugin-pwa'
    // import legacy from '@vitejs/plugin-legacy';
    
    export default defineConfig({
      plugins: [
        VitePWA({
          includeAssets: ['img/icons/favicon.png', 'img/icons/maskable_icon.png' ],
          manifest: {
            name: 'Final Countdown',
            start_url: "/",
            short_name: 'Final Countdown',
            description: 'Awesome countdown App',
            theme_color: '#031c36',
            icons: [
              {
                src: 'img/icons/chrome192.png',
                sizes: '192x192',
                type: 'image/png'
              },
              {
                src: 'img/icons/chrome512.png',
                sizes: '512x512',
                type: 'image/png'
              },
              {
                src: 'img/icons/chrome512.png',
                sizes: '512x512',
                type: 'image/png',
                purpose: 'any maskable'
              }
            ]
          }
        })
          
      ],
      build: {
        rollupOptions: {
          input: {
            main: resolve(__dirname, 'index.html'),
            about: resolve(__dirname, 'html/about.html'),
[...]

PR #397, opened 2022-10-15
feat: allow use assets in webmanifest icons

### LINKED ISSUE #211: Auto pull in references icons in manifest config
I was wondering if it would be possible to have any icons that are referenced to the relative file path in the vite project would then be picked up as a static asset like referencing a static asset in source code. Then the generated manifest would point to the generated icon i.e. 

set the below config

```js
VitePWA({
  manifest: {
    icons: [{
      src: './src/assets/images/android-chrome-192x192.png'
      sizes: '192x182',
      type: 'image/png'
    }]
  }
})
```

The plugin then will generate the manifest referencing the filepath generated by vite similar to how vite does static asset handling. With my setup of assetsDir: 'static' that  means I would expect the generated filepath to be something like `/static/android-chrome-192x192.cded8a63.png. 

Currently I have to have any of the above icons in my public directory.

userquin:
It works if referenced via static import or directly in src attribute: example on vue-router on `src/App.vue`: check the docs here https://deploy-preview-397--vite-plugin-pwa-legacy.netlify.app/guide/static-assets.html#reusing-src-assets-images

The docs will go to old docs and the new ones, this PR on draft until finished.

On dev server with dev options enabled it just works, Vite will resolve the `src/assets/xxx`  asset.

closes #211 and #396


userquin:
@antfu maybe we can have some problem with this on `vite-ssg`, since we're using it without the `bundle` param, I'll try to check it tmr on vitesse template

userquin:
I need to test all integrations, a fast review just checking source code:
- `vite-ssg`: no problem, build module just regenerates the service worker
- `îles`: should check if it is required, we're regenerating the webmanifest (https://github.com/ElMassimo/iles/blob/main/packages/pwa/src/pwa.ts#L156)
- `VitePress`: no problem
- `SvelteKit`: no problem
- `Astro`: no problem

userquin:
I Will try to emit the icon if missing fron the bundle, I think we shouldn't add this behavior.

If ppl Just want to add the icons to the assets folder instead public folder and those icons not being used (vía static import or used directly in the src attr) we need to do here the work done by Vite, no makes sense.

PR #401, opened 2022-10-28
docs: update react example deps

pure-js:
Hey!
@vitejs/plugin-react-refresh has been deprecated - https://www.npmjs.com/package/@vitejs/plugin-react-refresh
We can replace it with https://www.npmjs.com/package/@vitejs/plugin-react
What do you think about it?


userquin:
closing this, react example has been updated

PR #410, opened 2022-11-08
feat: split client type definitions

### LINKED ISSUE #409: Split type definitions for unexpected wrong type when monorepo
I have a monorepo with fifty projects. The monorepo include some vue and some react projects.
I need to avoid use React type into vue project.

https://github.com/vite-pwa/vite-plugin-pwa/blob/main/client.d.ts will effect this problem.
We need to split it.

userquin:
@yoyo930021 can you check with a local copy?

You will need to change `d.ts` or `tsconfig.json` to use:
```ts 
/// <reference types="vite-plugin-pwa/react" />
```
or (`tsconfig.json`)
```json
"types": ["vite-plugin-pwa/react"]
```

EDIT: your packages are using `vite-plugin-pwa/client`, I just move each fw to its own dts and remove all fw definitions from `client.d.ts`.

closes #409

### DIFF
--- a/preact.d.ts
+++ b/preact.d.ts
@@ -1,4 +1,8 @@
-declare module 'virtual:pwa-register' {
+declare module 'virtual:pwa-register/preact' {
+  // eslint-disable-next-line @typescript-eslint/prefer-ts-expect-error
+  // @ts-ignore ignore when preact/hooks is not installed
+  import type { StateUpdater } from 'preact/hooks'
+
   export interface RegisterSWOptions {
     immediate?: boolean
     onNeedRefresh?: () => void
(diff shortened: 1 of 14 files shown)


yoyo930021:
Looks great. but it is a breaking change. maybe we can keep client.d.ts content.

preact.d.ts
```diff
@@ -0,0 +1,32 @@
+declare module 'virtual:pwa-register/preact' {
+  // eslint-disable-next-line @typescript-eslint/prefer-ts-expect-error
+  // @ts-ignore ignore when preact/hooks is not installed
+  import type { StateUpdater } from 'preact/hooks'
+
+  export interface RegisterSWOptions {
```
antfu, line 6: Can we somehow share this interface?

userquin: @yoyo930021 @antfu can you check #486

userquin:
superseded by #486

PR #413, opened 2022-11-13
feat: support rollup assets file names

userquin:
By default `dontCacheBustURLsMatching` is using Vite assets file names (`<name>.<hash>.<ext>`) and  it will change to align it with Rollup default assets naming (`<name>-<hash>.<ext>`): https://github.com/vitejs/vite/pull/10878.

This PR just changes the regex to support both layouts.

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -85,7 +85,7 @@ export async function resolveOptions(options: Partial<VitePWAOptions>, viteConfi
     swSrc,
     swDest,
     globDirectory: outDirRoot,
-    dontCacheBustURLsMatching: /[.-].[a-f0-9]{8}\./,
+    dontCacheBustURLsMatching: /[.-][a-f0-9]{8}\./,
     injectionPoint: 'self.__WB_MANIFEST',
   }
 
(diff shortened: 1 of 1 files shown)


src/options.ts
```diff
@@ -76,7 +76,7 @@ export async function resolveOptions(options: Partial<VitePWAOptions>, viteConfi
     globDirectory: outDirRoot,
     offlineGoogleAnalytics: false,
     cleanupOutdatedCaches: true,
-    dontCacheBustURLsMatching: /\.[a-f0-9]{8}\./,
+    dontCacheBustURLsMatching: /[.-][a-f0-9]{8}\./,
```
benmccann, line 79: I think the `.` here needs to be escaped?

userquin: IIRC, in a character class (square brackets) any character except `^`, `-`, `]` or `\` is a literal, and `-` require escaping only if not the first nor last one.

userquin: About the dot, IntelliJ showing a redundant escaping warning:

imagen

PR #428, opened 2022-12-17
fix: prompt for update logic

userquin:
With current version multiple clients get auto updated once the service worker receives the first update: with logic included on PR https://github.com/vite-pwa/vite-plugin-pwa/pull/394

With current changes, it is working properly, also tested on Firefox.

There are a few combinations using `isUpdate` and `isExternal`: these 2 flags from `workbox-window` are defined as `boolean | undefined`, when `installed|activated` events fired, `undefined` values have special meaning.

As an example, with only one client:
- on first visit we receive `isExternal: false` with no `isUpdate`
- on update, we receive `isExternal: false` and `isUpdate: true` and `showSkipWaitingPrompt`  being called from `waiting` event.

Problems comes when we have a few clients, for example, 2 tabs in first browser instance and another browser instance:

On first visit we receive `isExternal: false` with no `isUpdate` only in the first one client visiting the app, other clients will not receive the event

On update it is more complex, since we can have a few combinations, here the tests I've made with previous setup (building between tests):

**Test 1**:
Reloading app in tab1:
- tab 1 receives `isExternal: false` and `isUpdate: true`
- tab 2 receives `isExternal: false` with no `isUpdate`
- second instance receives `isExternal: false` with no `isUpdate`

Click prompt reload button from tab1, all clients being updated, then build again.

**Test 2**:
Reloading browser instance 2:
- tab 1 receives `isExternal: false` and `isUpdate: true`
- tab 2 receives `isExternal: false` and `isUpdate: true`
- second instance receives `isExternal: false` and `isUpdate: true`

Click prompt reload button from browser instance 2, all clients being updated, then build again.

**Test 3**:
Reloading browser instance 2:
- tab 1 receives  `isExternal: false` with no `isUpdate`
- tab 2 receives  `isExternal: false` with no `isUpdate`
- second instance receives `isExternal: false` and `isUpdate: true`

Click prompt reload button from tab2 in browser instance 1, all clients being updated.

It seem the `isTernal` depends on the client that fires the update.

### DIFF
--- a/src/client/build/register.ts
+++ b/src/client/build/register.ts
@@ -25,7 +25,6 @@ export function registerSW(options: RegisterSWOptions = {}) {
   } = options
 
   let wb: import('workbox-window').Workbox | undefined
-  let registration: ServiceWorkerRegistration | undefined
   let registerPromise: Promise<void>
   let sendSkipWaitingMessage: () => Promise<void> | undefined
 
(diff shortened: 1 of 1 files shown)


userquin:
do not merge, I'm checking if we can switch event listener from `installed` to `activated`.

crunchwrap89:
Master @antfu can you please review this PR? <3

antfu (approved):

PR #440, opened 2023-01-06
docs: updates all links to docs to the new url

lukebatchelor:
Just a small update to fix all the broken links still pointing to the old docs url


userquin (approved):

userquin:
I don't remember if the deploy is still active, I think Anthony deactivated them...

PR #470, opened 2023-02-15
Create Laravel.md

fredpedersen:
Instructions for Laravel setup

### DIFF
--- a/docs/frameworks/Laravel.md
+++ b/docs/frameworks/Laravel.md
@@ -1,0 +1,60 @@
+---
+title: Laravel/Vite | Frameworks
+---
+
+# Laravel/Vite
+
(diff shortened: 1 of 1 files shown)


docs/frameworks/Laravel.md
```diff
@@ -0,0 +1,60 @@
...
+We will also make the plugin generate `registerSW.js` file which we can use to register the service worker:
+```ts
+VitePWA({
+  ...
+  registerSW: true,
+})
+```
+
+This generates a `registerSW.js` file which we should include along with the `manifest.webmanifest` file in any page which will be part of the PWA. You can edit the layout file to include:
+```
```
userquin, line 36: You can use both virtual modules, you dont need to add them manually.

Add how to use both virtual modules and move this as an opcional

fredpedersen: I never got that working I'm afraid. Feel free to edit.

userquin: upps, the docs here are deprecated, we should move this to docs repo

PR #483, opened 2023-03-17
fix: support "types" condition in "exports" field

sheremet-va:
If typescript uses "exports" field for resolution, it will ignore all other resolutions, so I have this error with the `bundler` `moduleResolution`:

```
error TS2688: Cannot find type definition file for 'vite-plugin-pwa/client'.
  The file is in the program because:
    Entry point of type library 'vite-plugin-pwa/client' specified in compilerOptions

  tsconfig.json:9:7
    9       "vite-plugin-pwa/client"
            ~~~~~~~~~~~~~~~~~~~~~~~~
    File is entry point of type library specified here.
```

https://arethetypeswrong.github.io

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,12 @@
       "require": "./dist/index.js",
       "import": "./dist/index.mjs"
     },
+    "./client": {
+      "types": "./client.d.ts"
+    },
+    "./info": {
+      "types": "./info.d.ts"
+    },
     "./*": "./*"
   },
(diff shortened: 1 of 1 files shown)


sheremet-va:
> We might need to also specify CJS and ESM entry for them

Are they meant to be imported directly? I don't see any actual `.js` files for these entry points.

antfu (approved):

PR #486, opened 2023-03-19
feat: virtual dts per framework

### LINKED ISSUE #409: Split type definitions for unexpected wrong type when monorepo

userquin:
supersedes #410
closes #409


userquin:
There are TS problems resolving `VanillaJS`, `React` and `Preact` register options when using `virtual:pwa-register` module, `registerSW/useRegisterSW` options resolved as `any`.

antfu (approved):

tresabhi:
I apologize for the delayed response. This solved the issue. However, now I do have to rewrite a good chunk of the startup code to use the `useRegisterSW` hook instead of `registerSW`.

PR #488, opened 2023-03-21
perf: calculate asset hash when required

### LINKED ISSUE #487: avoid calculate asset hashes when PWA disabled
When using pwa plugin in dev server without `devOptions` enabled, it tooks a while to calculate manifest hash entries when there is no need.

userquin:
closes #487


antfu (approved):
🔥

PR #572, opened 2023-09-29
Allow custom modes beyond production or development

mechanical-turk:
I noticed the necessity for this PR when I saw that my `sw-main.js` showed a different value for `import.meta.env.MODE` than my regular `main.js` after running `vite build --mode staging`. `sw-main.js` showed `production` while `main.js` showed `staging`.

This PR is an attempt to bring the `vite-plugin-pwa` configs a bit more inline with the [vite docs](https://vitejs.dev/guide/env-and-mode.html#modes) when it comes to modes. With this PR, modes can go beyond `production` and `development`. While `process.env.NODE_ENV` is a perfectly useful escape hatch, I believe defaulting the service worker's value for `mode` to the same value as the rest of the app will be more intuitive. We achieve this by:

a) No longer limiting the `mode` type to `production` | `development`, but also including `string`. We could also technically just do `string`, but including the literals could be a slightly better developer experience.

b) Actually passing the `options.mode` value down to `vite.build`.

c) Preserving the `process.node.env` escape hatch for backwards compatibility.

### DIFF
--- a/src/modules.ts
+++ b/src/modules.ts
@@ -99,8 +99,8 @@ export async function generateInjectManifest(options: ResolvedVitePWAOptions, vi
   await build({
     root: viteOptions.root,
     base: viteOptions.base,
-    // don't copy anything from public folder
     resolve: viteOptions.resolve,
+    // don't copy anything from public folder
     mode: options.mode,
     publicDir: false,
     build: {
--- a/src/modules.ts
+++ b/src/modules.ts
@@ -99,6 +99,7 @@ export async function generateInjectManifest(options: ResolvedVitePWAOptions, vi
   await build({
     root: viteOptions.root,
     base: viteOptions.base,
+    // don't copy anything from public folder
     resolve: viteOptions.resolve,
     mode: options.mode,
     publicDir: false,
--- a/src/modules.ts
+++ b/src/modules.ts
@@ -100,7 +100,7 @@ export async function generateInjectManifest(options: ResolvedVitePWAOptions, vi
     root: viteOptions.root,
     base: viteOptions.base,
     resolve: viteOptions.resolve,
-    // don't copy anything from public folder
+    mode: options.mode,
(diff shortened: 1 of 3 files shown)


userquin:
@mechanical-turk we need to do some testing since workbox-build modules using this env variable, we'll need to add some hint in the docs hinting the user to provide that variable in the env file.

userquin:
@mechanical-turk I'm doing some internal tests, I hadn't read this entry in the docs: https://vitejs.dev/guide/env-and-mode#node-env-and-modes (that's why thi PR was changed to `on hold`)

userquin:
So, using [this reproduction](https://github.com/userquin/vite-pwa-mode-repro--issue-571), running `vite build --mode <custom-mode>`  results:
- `development`: `process.env.NODE_ENV = production; mode = development`
- `preview`: `process.env.NODE_ENV = production; mode = preview`
- `production`: `process.env.NODE_ENV = production; mode = production`
- `development2`: `process.env.NODE_ENV = development; mode = development`
- `preview2`: `process.env.NODE_ENV = development; mode = preview`
- `production2`: `process.env.NODE_ENV = development; mode = production`

userquin:
@mechanical-turk 

After a few hours I finally fix the mode option, check previous vite link and the explanation in  #629, added 3 new options to allow do any combinationmwhen building the sw.

mechanical-turk:
@userquin thank you for putting in the work. excellent attention to detail and communication 🫡

userquin:
@mechanical-turk you can check new docs entry here: https://github.com/vite-pwa/docs/pull/122 and the page in the PR Preview here (open it in private browsing) https://deploy-preview-122--vite-pwa-org.netlify.app/guide/change-log.html#new-vite-build

userquin:
@mechanical-turk released `v0.18.0` and docs updated in inject manifest (guide section)

PR #590, opened 2023-11-04
refactor: use `order`/`handler` for transformIndexHtml

sapphi-red:
`enforce`/`transform` of transformIndexHtml hook is deprecated since Vite 4. Also an runtime warning is added in Vite 5.

This PR changes that to use `order`/`handler` instead for Vite 4+.


userquin (approved):
thx :tada:

PR #609, opened 2023-11-25
fix: remove sw cache storage entries when using self destroying option

### LINKED ISSUE #560: Self destroying service worker should clean up cache storage
When I deployed the self destroying service worker to my site , I noted that while the service worker did disappear, the saved file used to cache the site did not. While I probably should have read more into the sw, I do feel like its somewhat of an expected feature, so hopefully this gets implemented in some form.

userquin:
This PR also includes:
- check if the client is a `WindowClient` before calling `navigate`, it can be `Push/Sync`  worker (missing `navigate` method)
- bump to pnpm 8.11.0

closes #560

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "vite-plugin-pwa",
   "type": "module",
   "version": "0.17.0",
-  "packageManager": "pnpm@8.10.5",
+  "packageManager": "pnpm@8.11.0",
   "description": "Zero-config PWA for Vite",
   "author": "antfu <anthonyfu117@hotmail.com>",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


Doctor-wu:
@userquin I noticed that you deleted all the keys in cache-storage, maybe deleting keys generated by this plugin will be better?

userquin:
@Doctor-wu this plugin doesn't add any entry to the cache storage, the sw cache storage is per location (sw origin)

Doctor-wu:
@userquin users can use sw outside this plugin, and they can add other caches even if that is not recommended, but we can't prevent users from doing that. Thus maybe we can delete the cache literally declared in the workbox's `runtimeCaching`, so that we won't delete the cache outside this plugin

userquin:
@Doctor-wu we cannot detect user entries, you can use a custom one https://vite-pwa-org.netlify.app/guide/unregister-service-worker.html#custom-selfdestroying-service-worker

Doctor-wu:
@userquin Got you~, maybe we can inform users in docs that we will delete all the cache when they use self-destroying

Doctor-wu:
Sry, I only watched the danger block😂, and now that will be pretty good~

PR #612, opened 2023-11-28
Run configureStaticAssets when includeAssets option is true regardless of manifest option.

sgpinkus:
Fixes one line conditional enabling collection of static assets.

Currently `includeAssets` option has no effect unless `manifest` is also set. This makes it impossible to include additional assets while managing your own manifest file.


userquin (approved):
Nice catch!

PR #618, opened 2023-12-06
fix: add kolorist to dependencies

userquin:
We added the log to show workbox-build using kolorist, was there for the examples...


userquin:
closing this since it is inlined, there is a  problem  in pwa sveltekit plugin...

PR #623, opened 2023-12-14
feat(register): add on begin update hooks

Doctor-wu:
Add `onBeginUpdate` hooks when a new service worker is found and is being installed but not yet completed.


docs/frameworks/index.md
```diff
@@ -30,6 +30,7 @@ You can find all the `vite-plugin-pwa` virtual modules declarations in the follo
 declare module 'virtual:pwa-register' {
   export interface RegisterSWOptions {
     immediate?: boolean
+    onBeginUpdate?: () => void
```
userquin, line 33: remove docs changes, the docs folder here is deprecated (we need to add it to the docs repo): we have a redirection to new docs website

Doctor-wu: Copy

userquin (approved):
LGTM

thx

userquin:
can you add it to an example (don't use sveltekit, it is deprecated in this repo)?

Doctor-wu:
Done

src/client/build/register.ts
```diff
@@ -81,6 +82,8 @@ export function registerSW(options: RegisterSWOptions = {}) {
 
             onNeedRefresh?.()
           }
+          // Add an event listener to detect when the new service worker
+          // start to be installed but not yet installed.
```
userquin, line 86: can you remove these lines

userquin:
I'm going to do some tests (no idea when I'll have some free time) and maybe adding a new experimental option to use a new built-in `workbox-window` exposing the `install` event, `Workbox` repo seems dead.

examples/vue-router/src/ReloadPrompt.vue
```diff
@@ -25,27 +29,42 @@ const {
...
     }
   },
+  onBeginUpdate() {
+    console.log('Begin sw update message')
+    beginUpdateServiceWorker.value = true
+    lastBeginUpdateServiceWorkerTime.value = Date.now()
+  },
+})
+
+watch(computed(() => [offlineReady.value, needRefresh.value]), ([offlineReadyVal, needRefreshVal]) => {
```
userquin, line 39: why computed here? `watch(() => [offlineReady.value, needRefresh.value], ([offlineReadyVal, needRefreshVal]) => {`

Doctor-wu: For reactive with both offlineReady and needRefresh.
I guess maybe we shouldn't care about offineReady here?

userquin: `watch` isn't reactive?

Doctor-wu: Oh, I don't need to use computed here, just a pure function is OK

Doctor-wu: Solved

PR #629, opened 2023-12-23
feat!: new custom service worker build

### LINKED ISSUE #358: Some lines of code make the build have an export default
### Describe the bug

I'm writing a TypeScript file that is agnostic about whether it's run inside a service worker or inside a browser. The goal is that is for some reason the service worker fails to load, the browser can then just import the file instead.

Now this *almost* works. Except that when I add some lines of code, and it seems pretty arbitrary which lines of code will do it, the final build of the service worker will have an `export default N()` at the end. And of course a service worker is not allowed to have an export.

For example, if my agnostic file contains these lines of code:

```ts
const channel = new BroadcastChannel('memCache-commands')
// I'll add a line of code here in the next bit...

async function handleMessage(event: ExtendableMessageEvent | EMEvent): Promise<MemCacheMessage | void | ""> {
 // ...
}
```

Then the final build is fine. It starts with:
```js
var commonjsGlobal=typeof globalThis!="undefined"?globalThis:typeof window!="undefined"?window:typeof global!="undefined"?global:typeof self!="undefined"?self:{}
```

And it doesn't end with an `export default`. But if I change those few lines of code and add one more line, like so:

```ts
const channel = new BroadcastChannel('memCache-commands')
channel.addEventListener("message", handleMessage)

async function handleMessage(event: ExtendableMessageEvent | EMEvent): Promise<MemCacheMessage | void | ""> {
 // ...
}
```

Then suddenly the final build of the service worker starts like this:

```js
var O=(e,t)=>()=>(t||e((t={exports:{}}).exports,t),t.exports);var N=O((exports,module)=>{var commonjsGlobal=typeof globalThis!="undefined"?globalThis:typeof window!="undefined"?window:typeof global!="undefined"?global:typeof self!="undefined"?self:{}
```

And ends with:

```js
export default N();
```

This is not the only scenario where this happens. I've tried many variations to make this file environment agnostic and every time at the last few steps, it seems to break the final build. `npm run dev` still works fine btw, but that's because then it's building it as a module type instead of a classic type.

### Reproduction

Is my description above enough?

### System Info

```Shell
System:
    OS: macOS 12.4
    CPU: (8) arm64 Apple M1
    Memory: 114.17 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.15.1 - ~/.nvm/versions/node/v16.15.1/bin/node
    npm: 8.11.0 - ~/.nvm/versions/node/v16.15.1/bin/npm
  Browsers:
    Brave Browser: 89.1.22.71
    Chrome: 104.0.5112.79
    Safari: 15.5
```
```


### Used Package Manager

npm

### Validations

- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided reproduction is a [minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

## edit

I'm actually not sure if this is an issue of this plugin, or of rollup...

### LINKED ISSUE #566: vite sourcemap: 'hidden' doesn't quite work
If vite config is `build.sourcemap: 'hidden';`, the sourcemap is correctly generated for `service-worker.js.map` and `workbox.js.map`, but the sourceMappingURL is erroneously still included in the minified js files `service-worker.js` and `workbox.js`, even tho it should be hidden from inclusion at the end of the files.

It looks like vite-plugin-pwa coerces the value of build.sourcemap to a boolean, which means it isn't able to properly use the value.

### LINKED ISSUE #571: .env.[mode] files
Does VitePWA support Vite's `.env.[mode]` files?

https://vitejs.dev/guide/env-and-mode.html#env-files

I tried creating a very simple Vite react app (`yarn create vite my-react-app --template react-ts`) and installed VitePWA. My service worker has a log for `import.meta.env` to see what variables are passed in. I created a `.env.devAPI` file in the root of the project with an environment variable I want to specify. I added a command in my package.json `"build:devAPI": "vite --mode devAPI build"`. When run, in the main code I can see my added environment variable. However, in the service worker code I do not see it.

If instead I just add environment variables to a `.env` file, they do appear in the service worker. I guess this `.env.[mode]` is not supported by VitePWA.

### LINKED ISSUE #575: Output minification with injectManifest strategy
On both of my projects the output of the `filename` option is not minified (47Kb), comments are also kept.
I'm using default options with the `injectManifest` strategy and [TypeScript support](https://vite-pwa-org.netlify.app/guide/inject-manifest.html#typescript-support)

I've noticed that when I change the `injectManifest.rollupFormat` to `iife`, build output is minified (18Kb).

When I use the `generateSW` strategy, both files are minified (sw.js: 2Kb, workbox-xxxxxxxx.js: 14Kb)

vite: 4.4.1
vite-plugin-pwa: 0.16.5

### LINKED ISSUE #615: Importing a CommonJS node_module in a custom service worker turns it into an ES module
I've discovered that creating a custom service worker using the `injectManifest` strategy can lead to an issue when importing a `node_module` that is in the CommonJS format.  Specifically, the service worker will end up as an ES module, with an `export default`, which will result in an error when registered in the browser.  The built asset will look something like this:
```js
var t = (l, o) => () => (o || l((o = { exports: {} }).exports, o), o.exports);
var _ = t((r, e) => {
  // Your compiled code here
});
export default _();
```

I've created this minimal reproduction case [here](https://github.com/kgroat/Minimal-SW-ES-Repro).

I've discovered that adding this to the config gets rid of the default export, but I'm not sure if it has other side-effects:
```js
VitePWA({
  // the rest of your config up here...

  injectManifest: {
    rollupFormat: 'iife'
  }
})
```

### LINKED ISSUE #625: Can the build target option be added to the generateInjectManifest method?
I need to specify the browser compatibility version when building the serviceWorker file.

# Example
```javascript
// vite-plugin-pwa/src/modules.ts line 112
await build({
  root: viteOptions.root,
  base: viteOptions.base,
  resolve: viteOptions.resolve,
  // don't copy anything from public folder
  publicDir: false,
  build: {
    sourcemap: viteOptions.build.sourcemap,
    lib: {
      entry: options.swSrc,
      name: 'app',
      formats: [format],
    },
    rollupOptions: {
      ...rollupOptions,
      plugins,
      output: {
        entryFileNames: options.filename,
      },
    },
    outDir: options.outDir,
    emptyOutDir: false
+   target: viteOptions.build.target
// or
+   target: options.target
  },
```

### LINKED ISSUE #655: vite-wasm-plugin does not work with serviceworker
When trying to build a service worker that imports Webassembly I get the same error as if [vite-plugin-wasm](https://github.com/Menci/vite-plugin-wasm) is not there:
> "ESM integration proposal for Wasm" is not supported currently. Use vite-plugin-wasm or other community plugins to handle this. Alternatively, you can use `.wasm?init` or `.wasm?url`. See https://vitejs.dev/guide/features.html#webassembly for more details.

Building for developement does not throw an error but there is no serviceworker present.
Without the wasm import everything works just fine.

The config looks like this:
image

Am I configuring this wrong, is it a problem with VietPWA or is it a problem with the wasm plugin.

userquin:
After some digging, we found the problem in `esbuild` and the solution:
- change Vite build to use `.mjs` extension instead `.js`
- rename `.mjs` file to the original after the build
- rename also `.mjs.map` when sourcemap file generated

Thanks to @benmccann for filing the issue on Vite https://github.com/vitejs/vite/issues/15379 and @Rich-Harris for the solution on SvelteKit https://github.com/sveltejs/kit/pull/11400, this PR is a copy of the SvelteKit solution.

This PR also includes:
- include custom vite build module (vite.ts): extract logic to a new module
- include custom inject manifest strategy options for target, minify, sourcemap and dead code elimination: current version using mode option for minify and dead code elimination in workbox modules used in the sw; with new approach we can minify, add source map and use workbox logs in a production build (we can minify the code having sourcemap and workbox logs for example)
- deprecate `plugins` including new `buildPlugins` option for Rollup and Vite plugins: added an example and the corresponding test in vue-route example using Vite plugin with a simple virtual module

NOTE: added `lodash-es` to test it in Vite ecosystem-ci.

closes #566
closes #571
closes #575
closes #615
closes #625
closes #655

supersedes #572

### DIFF
--- a/examples/vanilla-ts-no-ip/src/custom-sw.ts
+++ b/examples/vanilla-ts-no-ip/src/custom-sw.ts
@@ -3,6 +3,9 @@ import { registerRoute } from 'workbox-routing'
 import { CacheFirst, NetworkFirst, StaleWhileRevalidate } from 'workbox-strategies'
 import { CacheableResponsePlugin } from 'workbox-cacheable-response'
 import { ExpirationPlugin } from 'workbox-expiration'
+import orderBy from 'lodash-es/orderBy.js'
+
+console.log(orderBy)
 
 declare let self: ServiceWorkerGlobalScope
 
--- a/src/modules.ts
+++ b/src/modules.ts
@@ -109,32 +109,63 @@ export async function generateInjectManifest(options: ResolvedVitePWAOptions, vi
 
   const { format, plugins, rollupOptions } = options.injectManifestRollupOptions
 
-  await build({
-    root: viteOptions.root,
-    base: viteOptions.base,
-    resolve: viteOptions.resolve,
-    // don't copy anything from public folder
-    publicDir: false,
-    build: {
-      sourcemap: viteOptions.build.sourcemap,
-      lib: {
-        entry: options.swSrc,
-        name: 'app',
-        formats: [format],
-      },
-      rollupOptions: {
-        ...rollupOptions,
-        plugins,
-        output: {
-          entryFileNames: options.filename,
+  if (format === 'iife') {
+    await build({
+      root: viteOptions.root,
+      base: viteOptions.base,
+      resolve: viteOptions.resolve,
+      mode: options.mode,
+      // don't copy anything from public folder
+      publicDir: false,
+      build: {
+        sourcemap: viteOptions.build.sourcemap,
+        lib: {
+          entry: options.swSrc,
+          name: 'app',
+          formats: [format],
         },
+        rollupOptions: {
+          ...rollupOptions,
+          plugins,
+          output: {
+            entryFileNames: options.filename,
+          },
+        },
+        outDir: options.outDir,
+        emptyOutDir: false,
       },
-      outDir: options.outDir,
-      emptyOutDir: false,
-    },
-    configFile: false,
-    define,
-  })
+      configFile: false,
+      define,
+    })
+  }
+  else {
+    await build({
+      root: viteOptions.root,
+      base: viteOptions.base,
+      resolve: viteOptions.resolve,
+      mode: options.mode,
+      // don't copy anything from public folder
+      publicDir: false,
+      build: {
+        modulePreload: false,
+        sourcemap: viteOptions.build.sourcemap,
+        rollupOptions: {
+          ...rollupOptions,
+          plugins,
+          input: options.swSrc,
+          output: {
+            entryFileNames: 'service-worker.mjs',
+            inlineDynamicImports: true,
+          },
+        },
+        outDir: options.outDir,
+        emptyOutDir: false,
+      },
+      configFile: false,
+      define,
+    })
+    await fs.rename(`${options.outDir}/service-worker.mjs`, `${options.outDir}/${options.filename}`)
+  }
 
   // don't force user to include injection point
   if (!options.injectManifest.injectionPoint)
--- a/src/modules.ts
+++ b/src/modules.ts
@@ -96,7 +96,7 @@ export async function generateInjectManifest(options: ResolvedVitePWAOptions, vi
 
   const [workbox, buildSW] = await Promise.all([
     loadWorkboxBuild(),
-    import('./vite').then(({ buildSW }) => buildSW),
+    import('./vite-build').then(({ buildSW }) => buildSW),
   ])
 
   await buildSW(options, viteOptions, workbox)
--- a/src/modules.ts
+++ b/src/modules.ts
@@ -109,7 +109,7 @@ export async function generateInjectManifest(options: ResolvedVitePWAOptions, vi
 
   const { format, plugins, rollupOptions } = options.injectManifestRollupOptions
 
-  const inlineConfig: InlineConfig = {
+  const inlineConfig = {
     root: viteOptions.root,
     base: viteOptions.base,
     resolve: viteOptions.resolve,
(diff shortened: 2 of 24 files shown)


src/modules.ts
```diff
@@ -109,32 +109,63 @@ export async function generateInjectManifest(options: ResolvedVitePWAOptions, vi
...
+      configFile: false,
+      define,
+    })
+  }
+  else {
+    await build({
+      root: viteOptions.root,
+      base: viteOptions.base,
+      resolve: viteOptions.resolve,
+      mode: options.mode,
```
antfu, line 146: I wonder if we could have a temporary object to put the share options, for future maintenance?

antfu (commented):
LGTM otherwise

src/modules.ts
userquin: sure

examples/vanilla-ts-no-ip/src/custom-sw.ts
```diff
@@ -3,6 +3,9 @@ import { registerRoute } from 'workbox-routing'
 import { CacheFirst, NetworkFirst, StaleWhileRevalidate } from 'workbox-strategies'
 import { CacheableResponsePlugin } from 'workbox-cacheable-response'
 import { ExpirationPlugin } from 'workbox-expiration'
+import orderBy from 'lodash-es/orderBy.js'
+
+console.log(orderBy)
```
Doctor-wu, line 8: This may not be needed?

userquin: It is a test for custom build esbuild, check vite issue

Doctor-wu: okay

src/vite-build.ts
```diff
@@ -0,0 +1,158 @@
...
+      emptyOutDir: false,
+    },
+    configFile: false,
+    define,
+  } satisfies InlineConfig
+
+  const swName = basename(options.swDest)
+  const swMjsName = swName.replace(/\.js$/, '.mjs')
+
+  if (format === 'iife') {
```
userquin, line 110: invert the condition

svihpinc:
@userquin thank you for resolving the linked issues. When can we expect this to be released as v0.18?

PR #634, opened 2023-12-31
feat: move to `@vite-pwa/workbox-window`

userquin:
This PR also includes `installing` event support for VanillaJS and all frameworks:
- VanillaJS: `onInstalling` and `onUpdateFound`
- Frameworks: same callbacks for VanillaJS and reactive `installingSW` and  `updatingSW`

**NOTE**: this PR should be included only in `v1.0.0` major version.

/cc @antfu I removed `ssr.noExternal` from main plugin, `@vite-pa/workbox-window` is ESM-first  with dual ESM/CJS package exports.

### DIFF
--- a/examples/preact-router/package.json
+++ b/examples/preact-router/package.json
@@ -54,6 +54,7 @@
   "devDependencies": {
     "@preact/preset-vite": "^2.7.0",
     "@rollup/plugin-replace": "^5.0.5",
+    "@vite-pwa/workbox-window": "^8.0.0",
     "https-localhost": "^4.7.1",
     "rimraf": "^5.0.5",
     "typescript": "^5.2.2",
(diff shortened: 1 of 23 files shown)


TheNoim:
Hey, it would be great if at least one of the two PRs (this one or #623) could actually land. Is there a particular reason why both PRs are stalled?

If there are some major roadblocks, it would be helpful to at least expose the Workbox instance to install the events temporarily yourself.

userquin:
We need to do more tests, rn the logic is wrong, workbox-window logic in vite pwa is wrong (we reverted a commit a few months ago in pwa plugin), we are also awaiting the new Google team, no idea if workbox repo Will be finally deprecated (2 years)

TheNoim:
> We need to do more tests, rn the logic is wrong, workbox-window logic in vite pwa is wrong (we reverted a commit a few months ago in pwa plugin), we are also awaiting the new Google team, no idea if workbox repo Will be finally deprecated (2 years)

Can we at least get https://github.com/vite-pwa/vite-plugin-pwa/pull/623 merged for now? The PR is smaller and would help for now.

userquin:
The logic is not 100% correct, we have some "race" condition when using multiple clients (tabs or browser instances); the `offlineReadyVal` logic is wrong since that value is true when the sw is installed and activated. The problem is the missing 'installing' event from workbox-window, it is NOT fired.

You can add the logic in your project if you want via `onRegisteredSW`.

PR #646, opened 2024-01-22
Add error handling to workbox import.

### LINKED ISSUE #645: Error handling for workbox import
I have a lot of users on older devices who run into issues with downloading the `workbox-window` module. Here is what shows up in Sentry, despite the file being available on my CDN:

CleanShot 2024-01-22 at 10 42 58@2x

Would it be acceptable to add try-catch behavior to the Workbox import here?

https://github.com/vite-pwa/vite-plugin-pwa/blob/main/src/client/build/register.ts#L40

cpojer:
I've seen the the dynamic import for workbox fail on older browsers, see #645. This PR adds basic error handling to ignore the workbox setup in such situations.

Fixes #645.

### DIFF
--- a/src/client/build/register.ts
+++ b/src/client/build/register.ts
@@ -37,7 +37,14 @@ export function registerSW(options: RegisterSWOptions = {}) {
 
   async function register() {
     if ('serviceWorker' in navigator) {
-      const { Workbox } = await import('workbox-window')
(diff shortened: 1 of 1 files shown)


cpojer:
Updated.

userquin:
Small refactor done:

imagen

userquin (approved):
thx

PR #659, opened 2024-02-14
Small fix to match info in types.ts

Bartek20:
Changed start_url to start_url + '?standalone=true'


userquin:
this can be configured from user land, why do we need this default?

Bartek20:
That was written in types.ts in @default value so I just changed it to match. If dafault should be clean base than let me change it in types.ts instead. :smile:

userquin (approved):

PR #686, opened 2024-03-15
fix(dev): inject register not working when disabling web manifest

### LINKED ISSUE #660: InjectRegister not working when disabling manifest
Even though i set injectRegister as null still it is auto Registering. How do i stop it ?

userquin:
closes #660

### DIFF
--- a/src/plugins/dev.ts
+++ b/src/plugins/dev.ts
@@ -32,7 +32,7 @@ export const swDevOptions = {
 export function DevPlugin(ctx: PWAPluginContext) {
   const transformIndexHtmlHandler = (html: string) => {
     const { options } = ctx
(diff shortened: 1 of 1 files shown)


rschristian:
Gave this a quick test but it doesn't seem to correct the issue, unless I'm being daft (perfectly possible).

userquin:
> Gave this a quick test but it doesn't seem to correct the issue, unless I'm being daft (perfectly possible).

I'm preparing some tests in my local, will take some time...

rschristian:
No worries or rush! Just thought I'd comment that

userquin:
I need to change also the `transformIndexHtml` in the dev plugin

userquin:
> No worries or rush! Just thought I'd comment that

With last commit should work: tested with `vanilla-ts-dev-options` example.

PR #769, opened 2024-10-15
Sw query params

### LINKED ISSUE #587: Query Parameters on the registration URL
Hi! Thank you for this wonderful plugin.

I'm working on something that would greatly benefit from being able to pass **query parameters to the service worker registration URL**. Something like, `/sw.js?version=1.0.3`. This way, within the service worker I am able to do `new URL(self.location).searchParams.get('version')`, and pass information to the service worker during registration. (See [SO](https://stackoverflow.com/a/46802958) answer)

## Example

One example of a use case would be that by sending a query parameter, I am able to convey to an incoming service worker information about the currently running application/service worker. Then, I can check whether the incoming service worker should force-install (via `self.skipWaiting()` and claim and refresh all clients in the `activate` event). I will be able to compare the version of the incoming service worker (let's say 2.0.0) with the current application (1.0.3 from the query parameter).

This is useful because it allows the incoming service worker to determine independently it's installation behavior, and we don't need to rely on currently running code (from the assets served by the existing service worker) to do the force update. This is fairly critical to my use case and I'm exploring what options are available.

```js
// sw.js

const version = `2.0.1`;

const url = new URL(self.location);
const currentVersion = url.searchParams.get('version');
	
const shouldAutoActivate = // check whether currentVersion -> version is a minor or major upgrade

self.addEventListener('install', () => {
	if (shouldAutoActivate) {
		self.skipWaiting();
	}
});

self.addEventListener('activate', () => {
	if (shouldAutoActivate) {
		return self.clients.matchAll()
		    .then(function(clients) {
				clients.forEach(client => client.navigate(client.url))
		    });
	}
});
```

## Alternative

Is there a way to send query parameters currently, that I am missing? I am using React, and there is not a way to use the `useRegisterSW()` utility hook to send query parameters, nor a way to set additional query parameters in the `VitePWAOptions`.

Iran-110:
### Description
For advanced use of service workers, such as versioning or complex caching, we sometimes need to send data to the service worker using search parameters like  `/sw.js?ver=1.2`.

This PR helps to solve this issue.

### Linked Issues
resolves #587.

### Additional Context
I use this pattern: https://github.com/vite-pwa/vite-plugin-pwa/issues/587#issuecomment-1818073650

### DIFF
--- a/docs/frameworks/index.md
+++ b/docs/frameworks/index.md
@@ -30,6 +30,7 @@ You can find all the `vite-plugin-pwa` virtual modules declarations in the follo
 declare module 'virtual:pwa-register' {
   export interface RegisterSWOptions {
     immediate?: boolean
+    searchQueryParams?: string | string[][] | URLSearchParams | Record<string, any>
     onNeedRefresh?: () => void
     onOfflineReady?: () => void
     /**
--- a/docs/frameworks/preact.md
+++ b/docs/frameworks/preact.md
@@ -23,6 +23,7 @@ declare module 'virtual:pwa-register/preact' {
 
   export interface RegisterSWOptions {
     immediate?: boolean
+    searchQueryParams?: string | string[][] | URLSearchParams | Record<string, any>
     onNeedRefresh?: () => void
     onOfflineReady?: () => void
     /**
--- a/docs/frameworks/react.md
+++ b/docs/frameworks/react.md
@@ -23,6 +23,7 @@ declare module 'virtual:pwa-register/react' {
 
   export interface RegisterSWOptions {
     immediate?: boolean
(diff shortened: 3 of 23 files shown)


userquin:
This should also work in dev mode, have you check it with some example? I'll try to review this PR next weekend (too busy, sorry )

Revert changes added to the docs, those changes should go to the docs repo here: https://github.com/vite-pwa/docs

Changes made to the examples won't pass Playwright tests (update PWA tests in the corresponding client-test folders or add an example only to any example not incuded in PW tests, check `test` script in root package.json): try running `pnpm test` from root folder, Playwright tests checking for the sw url in the browser page.

PR #784, opened 2024-11-13
feat(html): add head when missing from entry point

### LINKED ISSUE #782: `<head>` tag required for the manifest to be correctly injected
I spent several hours debugging why the `<link rel="manifest" href="/manifest.webmanifest">` tag wasn't injected into my app by `vite-plugin-pwa`.

Since I use [`unjs/unhead`](https://github.com/unjs/unhead) to manage all `<head>` tags, my `index.html` entry point didn't have a `<head>` tag:

```html
<!DOCTYPE html>
<html lang="en">
    <body>
        <div id="app"></div>
        <script type="module" src="/src/main.ts"></script>
    </body>
</html>
```

I realized that for `vite-plugin-pwa` to inject the manifest, an empty `<head>` tag is required:

```html
<!DOCTYPE html>
<html lang="en">
    <head>
    </head>
    <body>
        <div id="app"></div>
        <script type="module" src="/src/main.ts"></script>
    </body>
</html>
```

This should probably be documented somewhere.

userquin:
### Description

When there is no `head` in the html entry point, the sw and the web manifest may not be register: this PR checks for `</head>`  and `<body>` to add the head when required, showing a warning when missing both.

### Linked Issues

resolves #782

### Additional Context

---

> [!TIP]

### DIFF
--- a/src/html.ts
+++ b/src/html.ts
@@ -1,3 +1,4 @@
+import { yellow } from 'kolorist'
 import {
   DEV_PWA_ASSETS_NAME,
   DEV_READY_NAME,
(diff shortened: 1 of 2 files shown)


src/html.ts
```diff
@@ -25,20 +26,36 @@ navigator.serviceWorker.register('${path}', { scope: '${options.scope}' })
 }`.replace(/\n/g, '')
 }
 
+export function checkForHtmlHead(html: string) {
+  if (!html.includes('</head>')) {
+    if (!html.includes('<body>')) {
+      console.warn([
+        '',
+        yellow('PWA WARNING:'),
+        '</head> and <body> not found in the html, the service worker and web manifest will not be injected.',
```
userquin, line 35: replace this message with:
`</head> and <body> not found in the html, the service worker and the web manifest may not be injected.`

src/html.ts
```diff
@@ -25,20 +26,36 @@ navigator.serviceWorker.register('${path}', { scope: '${options.scope}' })
 }`.replace(/\n/g, '')
 }
 
+export function checkForHtmlHead(html: string) {
+  if (!html.includes('</head>')) {
```
userquin, line 30: should we also check for `<head />`? This will require to parse the html: Vite doesn't add the entries if `head` is missing and returning the tag array instead here

PR #788, opened 2024-11-18
chore: Remove dead code

piotr-cz:
### Description

At given point `event.isExternal` has already been evaluated to be `undefined`

https://github.com/vite-pwa/vite-plugin-pwa/blob/bee73adb56f3f08434981c022ecd5657d040ed55/src/client/build/register.ts#L94

so it's value cannot be truthy.

### Linked Issues

none

### Additional Context

I'm just trying to understand how the client-side logic works

### DIFF
--- a/src/client/build/register.ts
+++ b/src/client/build/register.ts
@@ -98,10 +98,7 @@ export function registerSW(options: RegisterSWOptions = {}) {
                   !onNeedRefreshCalled && onOfflineReady?.()
               }
               else {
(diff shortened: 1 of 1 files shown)


userquin:
The logic is in worbox window,  it is not documented and ot os hard to test, yoi need 2 or more tbs and 2 browser instances. Then fire the update from the tabs and from the other instance: depending who installed/activated the new sw the flag May change.

PR #799, opened 2024-11-27
fix: return async function from `registerSW` in dev mode

piotr-cz:
### Description

The signature for `registerSW` function should be same for both *production* and *development*:

```ts
function registerSW(options: RegisterOptions = {}): (_reloadPage?: boolean) => Promise<void>
```

### Linked Issues

None

### Additional Context

When trying to update service worker in *development*, following code throws an error:

```ts
import { registerSW } from 'virtual:pwa-register'

const updateServiceWorker = registerSW()

updateServiceWorker.then(() => {})
// Uncaught TypeError: Cannot read properties of undefined (reading 'then')
```


userquin (approved):
thx ❤️

PR #815, opened 2025-01-22
Allow only generating webmanifest (without service worker) as strategy

jnoordsij:
### Description

This PR aims to add a very simple `webManifestOnly` strategy to the plugin, which does not generate a service worker at all. This is useful for very minimal setups, which want to generate just a webmanifest file (possibly with additional assets generated through the assets-generator plugin).

### Linked Issues

N/A

### Additional Context

---

> [!TIP]

### DIFF
--- a/docs/guide/service-worker-strategies-and-behaviors.md
+++ b/docs/guide/service-worker-strategies-and-behaviors.md
@@ -8,9 +8,10 @@ A service worker strategy is related to how the `vite-plugin-pwa` plugin will ge
 
 ## Service Worker Strategies
 
-As we mention in [Configuring vite-plugin-pwa](/guide/#configuring-vite-plugin-pwa) section, `vite-plugin-pwa` plugin will use `workbox-build` node library to generate your service worker. There are 2 available strategies, `generateSW` and `injectManifest`:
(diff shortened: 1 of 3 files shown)


userquin:
To check the webmanifest you only need to add the plugin with empty options (the plugin will add some entries from the package.json), just enable dev options and test it in private browing.

I think this can be done also adding a manifest.json to public folder and registering the manifest link in the head, adding also this entry to the manifest.json file, and disabling the manifest (`manifest: false`) in the pwa options:
```json
{
  "$schema": "https://json.schemastore.org/web-manifest-combined.json"
}
```

The problem is about checking the new strategy here and all integrations when can be done on the user land, right now I'm very busy but will keep this PR open, be patient :pray:.

userquin:
I need to review this, maybe on next version, not sure yet if we are going to move vite-plugin-pwa to a new repo: right now I'm very busy working on workbox fork, but this will be a priority: we need to review the scope of the workbox fork.

netux:
:+1:ing this.

I tried this as a patch on top of v1.3.0 and it seems this crashes the build when using `@vite-pwa/sveltekit` v1.1.0.
I didn't feel like investigating further because I live with the sw.js and workbox .js files being included but not used in the build for now. But just as a heads up.

PR #843, opened 2025-03-25
fix: avoid assigning to bundle object

sapphi-red:
### Description

[Assigning to `bundle` object is discouraged by rollup](https://rollupjs.org/plugin-development/#generatebundle:~:text=DANGER,this.emitFile.) and is not supported by rolldown.

This PR changes the code using assignment to use `this.emitFile` instead. This would make vite-plugin-pwa work with Rolldown powered Vite.

If the plugin context is not passed to `api.generateBundle`, it will still use the previous method for backward compat.

### Additional Context

---

> [!TIP]

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,4 +1,4 @@
-import type { OutputBundle } from 'rollup'
+import type { OutputBundle, PluginContext } from 'rollup'
 import type { PWAPluginContext } from './context'
 import type { ExtendManifestEntriesHook, VitePluginPWAAPI } from './types'
 import { existsSync } from 'node:fs'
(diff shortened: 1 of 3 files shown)


userquin:
Thx ❤️ , I need to include this in Vite 6 Env. Api pr

userquin (approved):

sapphi-red:
> I'm checking also if we need to add the emit for `FILE_SW_REGISTER` asset via plugin context, just a few lines below changes in the api module:

Ah, yes, that should be changed as well. I'll update the PR.

userquin:
@sapphi-red just a question about vite and rolldown: we've small problem importing stuff from rollup, we should import rollup types from vite, but `OutputAsset` used in the api.ts module here not being re-exported by vite 5.0.12 (we should also update src/types.ts), so using rolldown will require a breaking dropping support for older Vite versions.

imagen

userquin (approved):
thx ❤️

PR #868, opened 2025-06-27
fix: add Vite 7.0.0 support

hasanparasteh:
Description

This PR adds Vite 7.0.0 support by updating the vite peer dependency range.
It allows users of Vite 7.x to use this plugin without peer dependency warnings.
Linked Issues

N/A
Additional Context

No additional changes or tests were necessary, as this is a peer dependency range update only.
Please let me know if you'd like me to include documentation updates or further compatibility checks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
   },
   "peerDependencies": {
     "@vite-pwa/assets-generator": "^1.0.0",
-    "vite": "^3.1.0 || ^4.0.0 || ^5.0.0 || ^6.0.0",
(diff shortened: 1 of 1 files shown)


hasanparasteh:
The lockfile didn’t change because the project was already using a compatible Vite version (Vite 5) and adding ^7.0.0 in package.json only permits 7.x — it doesn’t auto-upgrade the installed version. I can run pnpm up vite to update the lockfile if you’d like me to include that in this PR. Let me know!

userquin (approved):
thx :heart: sorry for the delay :pray:

PR #880, opened 2025-08-11
fix: add origin to scope_extensions to comply with the spec and get rid of warning

CodingDive:
### Description

Warning in chrome 139.0.7258.66

`manifest.webmanifest:1 Manifest: scope_extensions entry ignored, required properties 'type' and 'origin' are missing.`

### Linked Issues

Fixes https://github.com/vite-pwa/vite-plugin-pwa/issues/879

### Additional Context

I made the property optional since 'origin' is the only allowed type and there shouldn't be a breaking type change for such a minor issue. Therefore, I don't yet see the value in documenting it. If you'd like me to update some docs, let me know!

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -641,8 +641,8 @@ export interface ManifestOptions {
   scope_extensions: {
     origin: string
     /**
-      * @default 'origin'
-      */
+     * @default 'origin'
+     */
     type?: StringLiteralUnion<ScopeExtensionsType> | ScopeExtensionsType
   }[]
 }
--- a/src/types.ts
+++ b/src/types.ts
@@ -458,10 +458,7 @@ interface Nothing {}
  */
 export type StringLiteralUnion<T extends U, U = string> = T | (U & Nothing)
 
-/**
- * @default 'origin'
- * @see https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/scope_extensions
- */
+
 export type ScopeExtensionsType = 'origin'
 
 /**
@@ -635,12 +632,17 @@ export interface ManifestOptions {
   edge_side_panel?: {
     preferred_width?: number
   }
+
   /**
    * @see https://github.com/WICG/manifest-incubations/blob/gh-pages/scope_extensions-explainer.md
+   * @see https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/scope_extensions
    * @default []
    */
   scope_extensions: {
     origin: string
+    /**
+      * @default 'origin'
+      */
     type?: StringLiteralUnion<ScopeExtensionsType> | ScopeExtensionsType
   }[]
 }
--- a/src/types.ts
+++ b/src/types.ts
@@ -459,6 +459,12 @@ interface Nothing {}
 export type StringLiteralUnion<T extends U, U = string> = T | (U & Nothing)
 
 /**
+ * @default 'origin'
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -635,6 +635,7 @@ export interface ManifestOptions {
    */
   scope_extensions: {
     origin: string
+    type?: 'origin'
```
userquin, line 638: add a type: `export type ScopeExtensionsType = 'origin'` then use `type?: StringLiteralUnion<ScopeExtensionsType> | ScopeExtensionsType`, this way we can add future values for the type and TS won't complain: `export type ScopeExtensionsType = 'origin' | '<other1>' | ....` without the need to explicitly set the values, rn if a new type is added you need to use `unknown as any`  or similar hacks.

userquin: `StringLiteralUnion` should be there in the `src/types.ts` module

userquin: add also a `@see https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/scope_extensions` to the jsdoc

userquin: add also a jsdoc to the type with `@default 'origin'`

src/types.ts
```diff
@@ -458,6 +458,12 @@ interface Nothing {}
  */
 export type StringLiteralUnion<T extends U, U = string> = T | (U & Nothing)
 
+/**
+ * @default 'origin'
+ * @see https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/scope_extensions
+ */
```
userquin, line 464: move this jsdoc to L644

src/types.ts
```diff
@@ -635,6 +641,7 @@ export interface ManifestOptions {
    */
   scope_extensions: {
```
userquin, line 642: should be (no jsdoc in the ScopeExtensionsType): 
```ts
export type ScopeExtensionsType = 'origin'
```

adding additional `@see` + default value in the entry:
```ts
/**
   * @see https://github.com/WICG/manifest-incubations/blob/gh-pages/scope_extensions-explainer.md
   * @see https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/scope_extensions
   */
  scope_extensions: {
    origin: string
    /**
      * @default 'origin'
      */
    type?: StringLiteralUnion<ScopeExtensionsType> | ScopeExtensionsType
  }[]

File: src/types.ts [hunk 4]
```diff
@@ -629,12 +632,18 @@ export interface ManifestOptions {
...
    * @see https://github.com/WICG/manifest-incubations/blob/gh-pages/scope_extensions-explainer.md
+   * @see https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/scope_extensions
    * @default []
    */
   scope_extensions: {
     origin: string
+    /**
+      * @default 'origin'
+      */
+    type?: StringLiteralUnion<ScopeExtensionsType> | ScopeExtensionsType
```
userquin, line 646: can you remove extra spaces here, in the default value before the `*` and before `*/`:
```ts
    /**
     * @default 'origin'
     */
    type?: StringLiteralUnion<ScopeExtensionsType> | ScopeExtensionsType
```

userquin:
I will try to release a patch version next weekend...

userquin (approved):
thx ❤️ , I need to check this PR on my local

PR #882, opened 2025-08-22
fix: skip generation if the build errored

sapphi-red:
### Description

[`closeBundle` hook](https://rollupjs.org/plugin-development/#closebundle) is called even if an error occurred during build. Since the service worker generation and assets generation was done in that hook, it was done even when an error occurred.

This PR fixes that by checking the error parameter which was added in rollup v4.35.0 (https://github.com/rollup/rollup/pull/5867).
Note that this is not supported by Rolldown yet (https://github.com/rolldown/rolldown/issues/4364).

### Additional Context

---

> [!TIP]

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "vite-plugin-pwa",
   "type": "module",
-  "version": "1.0.2",
+  "version": "1.0.3",
   "packageManager": "pnpm@10.13.1",
   "description": "Zero-config PWA for Vite",
   "author": "antfu <anthonyfu117@hotmail.com>",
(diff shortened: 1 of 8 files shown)


userquin:
yes, but this way it is not breaking and will also work with rolldown, once supported then we're done

src/plugins/build.ts
```diff
@@ -40,7 +40,10 @@ export function BuildPlugin(ctx: PWAPluginContext) {
     closeBundle: {
       sequential: true,
       order: ctx.userOptions?.integration?.closeBundleOrder,
-      async handler() {
+      async handler(error) {
+        if (error)
```
userquin, line 44: I guess it is better moving the fn arg to:

```ts
async handler() {
  if (arguments?.length > 0 && arguments[0])
    return
```
}

userquin: no need to add a new dependency

PR #906, opened 2025-11-21
fix: resolveDevDistFolder to use outDir

gm112:
### Description

When using the generateSW function in dev mode, it seems the `dev-dist` folder resolves to the root, which in my react-router v7 project for [ZFGC.com](https://github.com/ZeldaFanGameCentral/ZFGBB-React) makes the dev sw stop working(I also have my outDir set to dist/client). This change modifies the function to use the configured outDir. 

### Linked Issues

N/A

### Additional Context

This seems to be an issue if you set the outDir

P.S. Thank you for this amazing plugin, by the way! I've used it on both Nuxt and React based projects with no issues! \o/

---


userquin:
rr7 not supported here, remix pwa integration not working (big breaking changes), I need to finish rr7 integration, you can check this PR in my repo (will be moved to PWA Organization once finished):

https://github.com/userquin/vite-pwa-react-router/pull/3

PR #914, opened 2026-02-07
feat(client): add onNeedReload callback

tomrehnstrom:
We have `@tanstack/router` and i have monkey-patched navigation to pass the reloadDocument boolean on the next navigation when we know there is a new version available. Currently we have a patch file that implements this in React.

## What
Adds a new `onNeedReload` callback to the `registerSW()` options. It is invoked whenever the plugin would normally hard-reload the page after a service worker update takes control.

## Why
Some SPAs need to fully control when a reload happens (e.g. defer it until the next router navigation, show a custom confirmation UI, or preserve in-progress form state). Today the built-in client always calls `window.location.reload()`.

## Details
- Runtime: `src/client/build/register.ts` calls `onNeedReload()` (fallback to `window.location.reload()`) in both:
  - `registerType: 'autoUpdate'` on `activated` (update/external)
  - `registerType: 'prompt'` on `controlling` after `updateSW()` triggers `skipWaiting`
- Framework wrappers: forward `onNeedReload` through `virtual:pwa-register/{react,vue,svelte,solid,preact}`.
- Types: updated `src/client/type.d.ts` and `types/index.d.ts`.
- Docs: updated guides and framework type declaration snippets to document `onNeedReload`.

## Notes
This preserves current behavior when `onNeedReload` is not provided.

### DIFF
--- a/docs/frameworks/index.md
+++ b/docs/frameworks/index.md
@@ -30,6 +30,7 @@ You can find all the `vite-plugin-pwa` virtual modules declarations in the follo
 declare module 'virtual:pwa-register' {
   export interface RegisterSWOptions {
     immediate?: boolean
+    onNeedReload?: () => void
     onNeedRefresh?: () => void
     onOfflineReady?: () => void
     /**
--- a/docs/frameworks/preact.md
+++ b/docs/frameworks/preact.md
@@ -23,6 +23,7 @@ declare module 'virtual:pwa-register/preact' {
 
   export interface RegisterSWOptions {
     immediate?: boolean
+    onNeedReload?: () => void
     onNeedRefresh?: () => void
     onOfflineReady?: () => void
     /**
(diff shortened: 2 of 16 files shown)


mikew:
Isn't this already possible by setting a flag in when `needRefresh` is `true` and then calling `updateServiceWorker(false)` and `window.location.href = nextLocation` when the navigation happens?

tomrehnstrom:
> Isn't this already possible by setting a flag in when `needRefresh` is `true` and then calling `updateServiceWorker(false)` and `window.location.href = nextLocation` when the navigation happens?

`updateServiceWorker(_reloadPage = true)`

The boolean argument is not used and it always reloads the page. I tried this solution at first but calling `updateServiceWorker` always reloads the page

Perhaps actually respecting the parameter is the better solution would be making that change

@mikew

userquin (approved):
Thx ❤️ 

You should change the docs at docs repo not here:

https://github.com/vite-pwa/docs

PR #930, opened 2026-05-05
docs(astro): fix grammar in Astro WIP placeholder

cipherfoxie:
Trivial grammar fix: \`Will coming soon\` → \`Coming soon\`.

\`will\` takes a bare infinitive (\`will come\`), not a present participle (\`coming\`). The simplest correction is to drop the modal verb and keep just \`Coming soon\`, which matches similar placeholders elsewhere in the docs.

### DIFF
--- a/docs/frameworks/astro.md
+++ b/docs/frameworks/astro.md
@@ -6,5 +6,5 @@ next: Getting Started | Examples
 # Astro
 
 ::: info WIP
-Will coming soon.
+Coming soon.
 :::


userquin:
Wrong repo, docs moved to its own repository here:
https://github.com/vite-pwa/docs

cipherfoxie:
Thanks for the pointer @userquin — apologies for the wrong-repo target. I'll re-submit the same fix on `vite-pwa/docs` since `deployment/aws.md` carries the identical "Will coming soon" placeholder there.

One small observation while I'm here: the orphan `docs/frameworks/astro.md` (and `iles.md`, both <12 lines) in this repo are still served via the netlify deploy preview, so search-engine landings could end up here instead of the canonical docs. Happy to open a follow-up PR removing them (or making them static redirects) if that fits your maintenance preferences. Just let me know.

PR #931, opened 2026-05-08
Fix: remove the `event.isUpdate` check in controlling handler

### LINKED ISSUE #789: Bug: Prompt for update doesn't reload page on first update
Steps to reproduce:

1. `git clone https://github.com/vite-pwa/vite-plugin-pwa.git`
1. Run examples as in [docs](https://vite-pwa-org.netlify.app/examples/preact.html)
1. Answer questions:
   - Select a framework: » preact
   - Select a strategy: » injectManifest
   - Select a behavior: » Prompt for update
   - Enable periodic SW updates? yes
   - Unregister SW? no
1. Open `https://localhost` in Chrome (optionally incognito mode)
1. In prompt *App ready to work offline*: click *Close*
1. Wait for 1 minute as suggested in docs
1. Repeat from step 2.
1. In prompt *New content available, click on reload button to update.* click *Reload*
1. Problem: Page does not reload

This happens only for first service worker update. For any subsequent updates, clicking on *Reload* reloads the page.
That's why reproduction should be run in incognito mode

Seems that the `controlling` event doesn't fire in this case:

https://github.com/vite-pwa/vite-plugin-pwa/blob/95142ebc588dea3d9376a2eb4affe086f19a7395/src/client/build/register.ts#L85-L88

piotr-cz:
### Description

The check for `event.isUpdate` is preventing the `onNeedReload()` callback from being executed on first update.

This event property indicates that it's an update to service worker that was available when workbox-window was registered.

In case of first update there wasn't a service worker at the time when Workbox has started, so it's value is `false`.

Full explanation and reproduction is in linked issue

### Linked Issues

fixes #789

### DIFF
--- a/src/client/build/register.ts
+++ b/src/client/build/register.ts
@@ -87,13 +87,11 @@ export function registerSW(options: RegisterSWOptions = {}) {
             // Assuming the user accepted the update, set up a listener
             // that will reload the page as soon as the previously waiting
             // service worker has taken control.
-            wb?.addEventListener('controlling', (event) => {
-              if (event.isUpdate) {
-                if (onNeedReload)
-                  onNeedReload()
-                else
-                  window.location.reload()
-              }
+            wb?.addEventListener('controlling', (_event) => {
+              if (onNeedReload)
+                onNeedReload()
+              else
+                window.location.reload()
             })
 
             onNeedRefresh?.()


userquin:
@piotr-cz I have sent you an invitation to my private monorepo, can you send me a dm at discord at antfu server (vite plugin pwa channel)?

https://discord.gg/rPacJqtkW

PR #932, opened 2026-05-08
fix(react): avoid duplicate SW registration in strict mode

### LINKED ISSUE #925: useRegisterSW (react) registers the SW twice in dev mode if strict mode is enabled
That also means that all the callbacks (onRegistered, onOfflineReady...) always fire twice.
Here's [the culprit](https://github.com/vite-pwa/vite-plugin-pwa/blob/main/src/client/build/react.ts#L20): react calls that useState cb twice in development.

cyphercodes:
### Description

Moves React `useRegisterSW` service worker registration out of the `useState` lazy initializer and into a guarded effect. The returned `updateServiceWorker` function stays stable and forwards to the registered updater via a ref, so React StrictMode development double-invocation does not register the service worker twice.

Added a focused Vitest coverage case that simulates StrictMode double-invoking state initializers/effects and verifies a single registration.

### Linked Issues

Fixes #925

### Additional Context

Verification:
- `pnpm run test-client`
- `pnpm run lint`
- `pnpm run build`
- `pnpm -C examples/react-router run run-build-claims && pnpm -C examples/react-router exec vitest run`
- `pnpm -C examples/react-router run run-build-sw-claims && SW=true pnpm -C examples/react-router exec vitest run`
- `git diff --check`

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,14 +100,15 @@
     "example:vue:build:cdn": "pnpm -C examples/vue-basic-cdn run build",
     "example:vue:start:cdn": "pnpm -C examples/vue-basic-cdn run start",
     "deploy": "nr build && nr docs:build",
+    "test-client": "vitest run test/*.test.ts",
     "test-vue": "pnpm -C examples/vue-router run test",
     "test-react": "pnpm -C examples/react-router run test",
     "test-preact": "pnpm -C examples/preact-router run test",
     "test-solid": "pnpm -C examples/solid-router run test",
     "test-svelte": "pnpm -C examples/svelte-routify run test",
     "test-typescript": "pnpm -C examples/vanilla-ts-no-ip run test",
-    "test": "nr test-vue && nr test-react && nr test-preact && nr test-solid && nr test-typescript",
-    "test:node20": "nr test-vue && nr test-react && nr test-preact && nr test-solid && nr test-svelte && nr test-typescript",
+    "test": "nr test-client && nr test-vue && nr test-react && nr test-preact && nr test-solid && nr test-typescript",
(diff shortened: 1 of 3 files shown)


userquin:
@cyphercodes 

I'm working on a new version of `vite-plugin-pwa` to add support also for `webpack 5` and `rspack` => `@vite-pwa/unplugin-pwa`.

The new `registerSW` becomes `void` and I have this using `useEffect` in a new virtual called `virtual:register-pwa/react-effect`, is this ok? Maybe I can just replace `react`  with this content.

In the meantime you can add a wrapper... I'm very busy finishing `workbox` fork and `@vite-pwa/unplugin-pwa`,, sorry :pray: 

```ts
import type { RegisterSWOptions } from '../types'
import { useEffect, useRef, useState } from 'react'
import { registerSW } from './register'

export type { RegisterSWOptions }

export function useRegisterSW(options: RegisterSWOptions = {}) {
  const {
    immediate = true,
    onNeedReload,
    onNeedRefresh,
    onOfflineReady,
    onRegisteredSW,
    onRegisterError,
  } = options

  const [needRefresh, setNeedRefresh] = useState(false)
  const [offlineReady, setOfflineReady] = useState(false)

  const registered = useRef(false)

  useEffect(() => {
    if (registered.current)
      return
    registered.current = true

    registerSW({
      immediate,
      onNeedReload,
      onOfflineReady() {
        setOfflineReady(true)
        onOfflineReady?.()
      },
      onNeedRefresh() {
        setNeedRefresh(true)
        onNeedRefresh?.()
      },
      onRegisteredSW,
      onRegisterError,
    })
  }, [])

  return {
    needRefresh: [needRefresh, setNeedRefresh],
    offlineReady: [offlineReady, setOfflineReady],
  }
}
```

PR #941, opened 2026-08-19
feat: allow defining color_scheme_dark in manifest

jrmajor:
### Description

Allows defining `color_scheme_dark` in manifest, as allowed by the standard: https://w3c.github.io/manifest/#color_scheme_dark-member.

### Linked Issues

No related issues.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -473,7 +473,7 @@ export interface IconResource {
   purpose?: StringLiteralUnion<IconPurpose> | IconPurpose[]
 }
 
-export interface ManifestColorScheme {
+export interface ColorSchemeDark {
   background_color?: string
   theme_color?: string
 }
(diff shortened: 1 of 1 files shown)


userquin:
Finally, I was reading the spec and there is also `*_localized` entries (I found only name).

src/types.ts
```diff
@@ -473,6 +473,11 @@ export interface IconResource {
   purpose?: StringLiteralUnion<IconPurpose> | IconPurpose[]
 }
 
+export interface ManifestColorScheme {
```
userquin: Rename this to ColorSchemeDark

jrmajor: Done.

src/types.ts
```diff
@@ -473,6 +473,11 @@ export interface IconResource {
   purpose?: StringLiteralUnion<IconPurpose> | IconPurpose[]
 }
 
+export interface ColorSchemeDark {
```
userquin, line 476: add link here to the spec, check other jsdocs annotations with `@see`

